PR #12, opened 2020-05-05
Add particle filter msgs to migration guide

naiveHobo:
Following https://github.com/ros-planning/navigation2/pull/1677


SteveMacenski:
FYI if you ever want to see what the website looks like in CI, you can go to circleci -> artifacts -> index.html and it'll pop up.

PR #13, opened 2020-05-08
Add DistanceController to plugin list

naiveHobo:
As per https://github.com/ros-planning/navigation2/pull/1699


SteveMacenski:
@naiveHobo please also add some reference to it in the migration guide from eloquent.

PR #20, opened 2020-06-01
Parameters descriptions

Marwan99:
This is a proposal for the parameters representation.

It probably makes more sense to have the parameters for each node in a separate page to avoid having them all crammed in single page.

Please let me know what you think.

### DIFF
--- a/configuration/configuring-the-bt.rst
+++ b/configuration/configuring-the-bt.rst
@@ -1,6 +1,6 @@
 .. _configuring_the_bt:
 
-Behavior-Tree-based Navigator
+Behavior-Tree Navigator
 #############################
 
 Source code on Github_.
--- a/configuration/configuring-the-bt.rst
+++ b/configuration/configuring-the-bt.rst
@@ -108,23 +108,23 @@ Example
         transform_tolerance: 0.1
         bt_xml_filename: replace/with/path/to/bt.xml
         plugin_lib_names: 
-        -nav2_compute_path_to_pose_action_bt_node
-        -nav2_follow_path_action_bt_node
-        -nav2_back_up_action_bt_node
-        -nav2_spin_action_bt_node
-        -nav2_wait_action_bt_node
-        -nav2_clear_costmap_service_bt_node
-        -nav2_is_stuck_condition_bt_node
-        -nav2_goal_reached_condition_bt_node
-        -nav2_initial_pose_received_condition_bt_node
-        -nav2_goal_updated_condition_bt_node
-        -nav2_reinitialize_global_localization_service_bt_node
-        -nav2_rate_controller_bt_node
-        -nav2_distance_controller_bt_node
-        -nav2_recovery_node_bt_node
-        -nav2_pipeline_sequence_bt_node
-        -nav2_round_robin_node_bt_node
-        -nav2_transform_available_condition_bt_node
+        - nav2_compute_path_to_pose_action_bt_node
+        - nav2_follow_path_action_bt_node
+        - nav2_back_up_action_bt_node
+        - nav2_spin_action_bt_node
+        - nav2_wait_action_bt_node
+        - nav2_clear_costmap_service_bt_node
+        - nav2_is_stuck_condition_bt_node
+        - nav2_goal_reached_condition_bt_node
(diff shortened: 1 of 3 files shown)


SteveMacenski:
I like the example at the end of the description, I think that can be helpful for a copy paste. I wonder if we should put it first or last though, what do you think?

On multiple pages: Do you think it would be better as `N` pages with no accordions or ` page with `N` accordions? I could go either way.  

Otherwise, I like it though with a couple tweaks:
- There's a bunch of spelling issues :smile: 
- The Example should be fully described with all the parameters, since this will be the only place probably that all parameters will be shown.
- I think you can just start with the first parameter, the name + "Paramaters" headers beforehand are a little redundant on the parameter page. 

@crdelsey for some reason, CI isn't showing up in the PRs but its being run https://84-240847415-gh.circle-artifacts.com/0/html/configuration/params/tunable-params.html. Any ideas?

SteveMacenski:
Thoughts on the multiple vs single page?

Marwan99:
I think multiple page will allow provide us with more space to add more content in the future. First example comes to mind is adding some explanation on each node like  http://wiki.ros.org/costmap_2d 

I am just afraid it will be too much if we have everything in the same page, making it harder for developers to find what they are looking and make it overwhelming for someone just starting.

Also accordions need to be expanded before being able to use CTRL+F, which is one more step.

What are your thoughts?

Marwan99:
Regarding the position of the example, I think it flows more naturally to have the information presented followed by an example(s) showing how to compile the information into the required format.

SteveMacenski:
Multiple pages are fine then. Order is OK then. Can you make this the template for what you want to do with multiple pages?

Marwan99:
If the format in the previous commit is fine (it is actually quite similar to the existing controller page), will go ahead and fully update the controller pages as well.

configuration/configuring-the-bt.rst
```diff
@@ -0,0 +1,133 @@
...
+
+    bt_navigator:
+      ros__parameters:
+        use_sim_time: true
+        global_frame: map
+        robot_base_frame: base_link
+        transform_tolerance: 0.1
+        bt_xml_filename: replace/with/path/to/bt.xml
+        plugin_lib_names: 
+        -nav2_compute_path_to_pose_action_bt_node
```
SteveMacenski, line 111: Doesn't yaml have a space after the `-` before the namespaced object?

configuration/configuring-the-bt.rst
```diff
@@ -0,0 +1,133 @@
+.. _configuring_the_bt:
+
+Behavior-Tree-based Navigator
```
SteveMacenski, line 3:
```suggestion
Behavior-Tree Navigator
```

Marwan99:
For the landing page, is there a diagram that represents how all the components interact together?
Something like this [diagram](http://wiki.ros.org/navigation/Tutorials/RobotSetup?action=AttachFile&do=view&target=overview_tf.png) from ROS1, but updated

SteveMacenski:
There's one on the main navigation ros org index page, but not sure how that helps with the landing page for the configuration guide. I was thinking maybe we do something like the accordians (big icons of some kind) that you hit to direct to the specific pages. But the list is also fine for now. I don't think we should over-optimize for this right now but if you had some thoughts and a quick prototype, might be worth it. If not, no worries.

Marwan99:
Did a quick google search but could not find something that could be tested quickly. So I am afraid I don't have any ideas for improving visuals atm.

configuration/params/tunable-params.rst
```diff
@@ -410,10 +410,116 @@ Hosts multiple recovery actions
...
     bt_navigator:
       ros__parameters:
-        bt_xml_filename: navigate_w_replanning_and_recovery.xml
         use_sim_time: true
+        global_frame: map
+        robot_base_frame: base_link
+        transform_tolerance: 0.1
+        bt_xml_filename: replace/with/path/to/bt.xml
+        plugin_lib_names: 
+        -nav2_compute_path_to_pose_action_bt_node
```
SteveMacenski, line 509: Spaces after lines here too

Issue #21, opened 2020-06-08
Updating the parameters in the config guidelines


Marwan99:
List of parameters with their state (https://github.com/ros-planning/navigation2/blob/master/doc/parameters/param_list.md):
- [x] bt_navigator
- [x] costmaps
- [x] DWB
- [x] waypoint follower 
- [x] controller_server
- [x] lifecycle_manager
- [x] map_server
- [x] planner_server
- [x] navfn_planner
- [x] recoveries
- [x] AMCL
- [x] Behavior Tree Nodes

SteveMacenski:
@Marwan99 I'm going to do the lifecycle manager and waypoint follower right now

SteveMacenski:
And controller server, because I had a few minutes to kill

Marwan99:
@SteveMacenski I will have a go at the bt nodes.

SteveMacenski:
My plan today was the planner server and navfn plugin.

SteveMacenski:
Also did mapserver

SteveMacenski:
Once we're done with all of them, we should remove the tunable parameters page since it will be unnecessary

SteveMacenski:
recoveries done... OK done for the day for real now

SteveMacenski:
I'll do AMCL tomorrow. You mind taking DWB and the costmaps?

SteveMacenski:
Eh, I'll do AMCL right now. Which would you rather do; DWB or costmaps? I'll do the other.

Marwan99:
I go with DWB.

SteveMacenski:
*phew* I wanted to do costmaps so that's a good fit :laughing:

SteveMacenski:
Sweet, we might be done with this by the end of the week. That would be great. I'll post on discourse about it once we're done.

SteveMacenski:
@Marwan99 when you're done, remember to delete the existing DWB pages and the "tuning the controller' page

PR #25, opened 2020-06-09
add bt xml config and plugins

Marwan99:
per #21


Marwan99:
@SteveMacenski Just want an okay from you to go ahead with this format. A more or less ready example to check is the Backup action.

SteveMacenski:
yeah this example looks great. I think you just need a couple sentences at the top of each page for what each of them does, and a bunch are missing examples. Otherwise, LGTM.

configuration/packages/configuring-bt-xml.rst
```diff
@@ -1,4 +1,4 @@
-.. _configuring_behavior_tree:
+.. _configuring_behavior_tree_xml:
 
 Configuring the behavior tree XML
```
SteveMacenski, line 3:
```suggestion
Behavior Tree XML Nodes
```

SteveMacenski: This title is a little long and not in the same theme as the others

Marwan99:
Still a couple more examples and descriptions need to be added. Will call it a day today and finish this tomorrow (technically it's already tomorrow my time xD).

SteveMacenski:
OK. Only thing I notice is that there's no parameters listed (e.g. non input ports, but actual ROS parameters that individual nodes use in the BT navigator node namespace). Listing the parameters in each BT node separately of port would be nice since a few have them I think.

For the cases where the input port is taken from the blackboard, I don't see that represented in anyway. I think `goal` and `path` are the only 2 though in compute path to pose so that's a small update

Marwan99:
I had a look at the param list and the only params I think are relevant here are the recoveries but I can see you adding them in #26. Can you please point me to the ones you are talking about?

SteveMacenski:
I did a run through of all of the BT nodes, these 2 have parameters from the BT nav

https://github.com/ros-planning/navigation2/blob/master/nav2_behavior_tree/plugins/condition/goal_reached_condition.cpp

https://github.com/ros-planning/navigation2/blob/master/nav2_behavior_tree/plugins/condition/distance_traveled_condition.cpp

I wonder if we should at minimum make a note that they use the BT navigator's value for these parameters, so they know it effects it

Marwan99:
A note as you mentioned with maybe a link to the bt navigator params might do it. I will also need to add a page for the distance travelled condition.

SteveMacenski:
Oh I didn't check if you included them all,  I assumed you had

Marwan99:
Well.. I was assuming everything was in the param file, sorry about that. Will go through the code and make sure we have everything.

SteveMacenski:
Did you add the 2 nodes with parameters yet? I think after that this is good to ship now that you added the missing nodes

Marwan99:
More things coming, give me a couple of minutes.

Marwan99:
Do we need to have `name` added as input port? 

Other than I think it is ready for review. Some of the conditions and the round robin action need descriptions and/or descriptions for their parameter, as those were missed in the param md file and didn't find documentation for them.

SteveMacenski:
What do you mean by `name`?

Marwan99:
Sorry `name` is just an attribute not a port, but it is optional https://www.behaviortree.dev/xml_format/.

Examples of `name` usage in a bt:
https://github.com/ros-planning/navigation2/blob/caea7fa4e68be18f6018796b983c8d008a8b492e/nav2_bt_navigator/behavior_trees/navigate_w_replanning_and_recovery.xml#L8-L13

SteveMacenski:
Ah ok. I think its fine to not include it, but add a note on the parent page of the BT node sub-pages that they all have this `name` attribute for identifying in logging. 

Which of the files did you need descriptions / examples for? If you enumerate them, I'll provide them.

Marwan99:
- Wait action (I think it needs improving)
- ReinitializeGlobalLocalization action (missing description)
- DistanceTraveled condition (missing description and distance param description)
- InitialPoseReceived condition (missing description)
- TimeExpired condition (missing description and seconds param description)
- RoundRobin control (missing description)

SteveMacenski:
Reinitialize description: Used to trigger global relocalization using AMCL in case of severe delocalization or kidnapped robot problem.

Distance Traveled: Node that returns success when a configurable distance has been traveled.
(the distance parameter is the distance it must travel before returning success (m) )

Initial Pose Received: Node that returns success when the initial pose is sent to AMCL via `/initial_pose`.

Time Expired: Node that returns success when a time duration has passed
(seconds is the time passed to return success (s) )
RoundRobin: Custom control flow node used to create a round-robbin behavior for children BT nodes.

SteveMacenski:
Let me know when I should look again

Marwan99:
I think its ready for reviewing.

SteveMacenski:
Can you send me a link to the index.html for this? I can't seem to find it

Marwan99:
https://164-240847415-gh.circle-artifacts.com/0/html/index.html

PR #33, opened 2020-06-23
Add updated plugin map parameter parsing mechanics

naiveHobo:
Signed-off-by: Sarthak Mittal <sarthakmittal2608@gmail.com>

As per: https://github.com/ros-planning/navigation2/issues/1791 and https://github.com/ros-planning/navigation2/pull/1808

### DIFF
--- a/configuration/packages/configuring-controller-server.rst
+++ b/configuration/packages/configuring-controller-server.rst
@@ -24,7 +24,7 @@ Parameters
   Description
     Frequency to run controller (Hz).
 
-:controller_plugin_ids:
+:controller_plugins:
 
   ============== ==============
   Type           Default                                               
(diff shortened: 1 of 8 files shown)


configuration/packages/configuring-controller-server.rst
```diff
@@ -35,16 +35,16 @@ Parameters
   Description
     List of mapped names for controller plugins for processing requests and parameters.
 
-:controller_plugin_types:
+:FollowPath.plugin:
```
SteveMacenski, line 38: This should be < controller plugin > or something - see other examples in the configuration section with user-defined namespaces

SteveMacenski: Ditto for all below

SteveMacenski: This one is hard because we need to define for all of these cases both a default and a generic parameter. I think we should stick with the generic parameter, but in the description, say that the default controller / planner / etc plugin namespace is `FollowPath`.

PR #46, opened 2020-07-14
Added clarification on footprint parameter syntax

Michael-Equi:
Related to https://github.com/ros-planning/navigation2/issues/1874

### DIFF
--- a/configuration/packages/configuring-costmaps.rst
+++ b/configuration/packages/configuring-costmaps.rst
@@ -44,7 +44,7 @@ Costmap2D ROS Parameters
   ============== =======
 
   Description
-    Ordered set of footprint points passed in as a string, must be closed set.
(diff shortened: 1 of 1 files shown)


configuration/packages/configuring-costmaps.rst
```diff
@@ -40,11 +40,11 @@ Costmap2D ROS Parameters
   ============== =======
   Type           Default
   -------------- -------
-  vector<double> {}   
+  vector<double> "[]"   
   ============== =======
 
   Description
-    Ordered set of footprint points, must be closed set.
+    Ordered set of footprint points passed in as a string, must be closed set.
```
SteveMacenski, line 47: Can you throw in an example with the spacing since that was part of the issue?

PR #63, opened 2020-08-04
Added map_topic documentation for amcl and static costmap layer

Michael-Equi:
For PR https://github.com/ros-planning/navigation2/pull/1910


SteveMacenski:
We'll revisit this once we deal with the main PR. But also add this to the migration guide that we've exposed these.

PR #65, opened 2020-08-06
map_saver does not exist, map_saver_cli does

mikeferguson:


SteveMacenski:
FYI there's also a new map saver server that can save maps live through a service interface, in case a one-off save isn't interesting to you (like me)

PR #81, opened 2020-09-01
add ubr to robots

mikeferguson:


SteveMacenski:
https://403-240847415-gh.circle-artifacts.com/0/html/about/robots.html the render doesn't show the image

PR #82, opened 2020-09-03
Add IsBatteryLow BT Condition node info

naiveHobo:
As per https://github.com/ros-planning/navigation2/pull/1974


configuration/packages/bt-plugins/conditions/IsBatteryLow.rst
```diff
@@ -0,0 +1,15 @@
+.. _bt_is_battery_low_condition:
```
SteveMacenski, line 1: Add to foxy to G-turtle migration guide.

SteveMacenski (approved):

PR #95, opened 2020-10-09
adding controller server plugins + reformatting waypoint follower's

SteveMacenski:


SteveMacenski:
@jediofgever  for future reference, the table is for implementations of specific plugins available, not the general  class of plugin definition. I updated your waypoint task executor table to include the wait plugin implementation :+1: My bad for not catching that.

PR #120, opened 2020-12-06
Standardize time units per REP-103

simutisernestas:
synch with https://github.com/ros-planning/navigation2/pull/2106


SteveMacenski (approved):
LGTM, waiting on the nav2 PR Ci to pass so we can merge these together

Issue #123, opened 2020-12-15
Tutorial: Speed restricted zones


SteveMacenski:
@AlexeyMerzlyakov

AlexeyMerzlyakov:
Thanks for noting :+1:  I'll take care of it.

PR #124, opened 2020-12-19
Add tutorial for Speed Filter

AlexeyMerzlyakov:
Adding the tutorial for Speed Filter (https://github.com/ros-planning/navigation.ros.org/issues/123 task)


AlexeyMerzlyakov:
Added the initial version of Speed Filter tutorial. Everything is made, except demonstration video. After all texts will be approved, I will shoot and attach the video in the beginning of this tutorial.

tutorials/docs/navigation2_with_speed_filter.rst
```diff
@@ -0,0 +1,327 @@
...
+
+It is assumed that ROS 2, Gazebo and TurtleBot3 packages are installed or built locally. Please make sure that the Navigation2 project is also built locally as it was made in :ref:`build-instructions`.
+
+Tutorial Steps
+==============
+
+1. Prepare filter mask
+----------------------
+
+As was written in :ref:`concepts`, any Costmap Filter (including Speed Filter) are reading the data marked in a filter mask file. All information about filter masks, their types, detailed structure and how to make a new one is written in a :ref:`navigation2_with_keepout_filter` tutorial at "1. Prepare filter maks" chapter. The principal of drawing filter mask for Speed Filter is the same as for Keepout Filter, except that ``OccupancyGrid`` mask values have another meaning: these values are encoded speed limits for the areas corresponding to the cell on map.
```
SteveMacenski, line 34:
```suggestion
As was written in :ref:`concepts`, any Costmap Filter (including Speed Filter) is reading the data marked in a filter mask file. All information about filter masks, their types, detailed structure and how to make a new one is written in a :ref:`navigation2_with_keepout_filter` tutorial at ``1. Prepare filter masks`` chapter. The principal of drawing the filter mask for Speed Filter is the same as for Keepout Filter (to annotate a map with the requested zones), except that ``OccupancyGrid`` mask values have another meaning: these values are encoded speed limits for the areas corresponding to the cell on map.
```

tutorials/docs/navigation2_with_speed_filter.rst
```diff
@@ -0,0 +1,327 @@
...
+
+ - Speed limit expressed in a percent from maximum robot speed.
+ - Speed limit expressed in absolute values (e.g. in ``m/s``).
+
+The meaning used by Speed Filter is being read from ``nav2_msgs/CostmapFilterInfo`` messages.
+In this tutorial we will use the first type of speed restriction expressed in a percent from maximum robot speed.
+
+.. note::
+
+  For speed restriction expressed in a percent, ``speed_limit`` will be used exactly as a percent belonging to ``[0..100]`` range, but not as a share from ``[0.0..1.0]`` range.
```
SteveMacenski, line 57:
```suggestion
  For speed restriction expressed in a percent, ``speed_limit`` will be used exactly as a percent belonging to ``[0..100]`` range, not ``[0.0..1.0]`` range.
```

tutorials/docs/navigation2_with_speed_filter.rst
```diff
@@ -0,0 +1,327 @@
...
+
+Create a new image with a PGM/PNG/BMP format: copy ``turtlebot3_world.pgm`` main map which will be used in a world simulation from a navigation2 repository to a new ``speed_mask.pgm`` file. Open ``speed_mask.pgm`` in your favourite raster graphics editor and fill speed restricted areas with grey colors. In our example darker colors will indicate areas with higher speed restriction:
+
+.. image:: images/Navigation2_with_Speed_Filter/drawing_speed_mask.png
+    :width: 500px
+
+Area "A" is filled with ``40%`` gray color, area "B" - with ``70%`` gray, what should mean that speed restriction will take ``100% - 40% = 60%`` in area "A" and ``100% - 70% = 30%`` in area "B" from maximum speed value allowed for this robot.
```
SteveMacenski, line 64:
```suggestion
Area "A" is filled with ``40%`` gray color, area "B" - with ``70%`` gray, that means that speed restriction will take ``100% - 40% = 60%`` in area "A" and ``100% - 70% = 30%`` in area "B" from maximum speed value allowed for this robot.
```

tutorials/docs/navigation2_with_speed_filter.rst
```diff
@@ -0,0 +1,327 @@
...
+We will use ``scale`` map mode with no thresholds. In this mode darker colors will have higher ``OccupancyGrid`` values. E.g. for area "B" with ``70%`` of gray ``OccupancyGrid`` data will be equal to ``70``. So in order to hit the target (darker colors indicate areas with higher speed restriction), we need to choose ``base = 100.0`` and ``multiplier = -1.0``. This will reverse the scale ``OccupancyGrid`` values to a desired one.
```
SteveMacenski, line 65:
```suggestion
We will use ``scale`` map mode with no thresholds. In this mode darker colors will have higher ``OccupancyGrid`` values. E.g. for area "B" with ``70%`` of gray ``OccupancyGrid`` data will be equal to ``70``. So in order to hit the target, we need to choose ``base = 100.0`` and ``multiplier = -1.0``. This will reverse the scale ``OccupancyGrid`` values to a desired one.
```

You already mentioned in the sentence prior

tutorials/docs/navigation2_with_speed_filter.rst
```diff
@@ -0,0 +1,327 @@
...
+  image: speed_mask.pgm
+
+  mode: trinary
+  ->
+  mode: scale
+
+  occupied_thresh: 0.65
+  free_thresh: 0.196
+  ->
+  occupied_thresh: 1.0
```
SteveMacenski, line 90: are these 2 thresholds even used in this situation? Is it important to change at all?

tutorials/docs/navigation2_with_speed_filter.rst
```diff
@@ -0,0 +1,327 @@
...
+  free_thresh: 0.0
+
+Save ``speed_mask.yaml`` and the new filter mask is ready to use.
+
+.. note::
+
+  World map itself and filter mask could have different sizes, origin and resolution which might be useful e.g. for cases when filter mask is covering smaller areas on maps or when one filter mask is used repeatedly many times (like annotating a speed restricted area for same shape rooms in the hotel). For this case, you need to correct ``resolution`` and ``origin`` fields in YAML as well so that the filter mask is correctly laid on top of the original map.
```
SteveMacenski, line 97:
```suggestion
  World map itself and filter mask could have different sizes, origin and resolution which might be useful (e.g. for cases when filter mask is covering smaller areas on maps or when one filter mask is used repeatedly many times, like annotating a speed restricted area for same shape rooms in the hotel). For this case, you need to correct ``resolution`` and ``origin`` fields in YAML as well so that the filter mask is correctly laid on top of the original map. This example shows using the main map as a base, but that is not required.
```

tutorials/docs/navigation2_with_speed_filter.rst
```diff
@@ -0,0 +1,327 @@
...
+
+.. note::
+
+  Another important note is that since Costmap2D does not support orientation, the last third "yaw" component of the ``origin`` vector should be equal to zero. For example: ``origin: [1.25, -5.18, 0.0]``.
```
SteveMacenski, line 101: Remove this as a note and add it as a sentence in the yaml file modification section above

tutorials/docs/navigation2_with_speed_filter.rst
```diff
@@ -0,0 +1,327 @@
...
+      type: 1
+      filter_info_topic: "/costmap_filter_info"
+      mask_topic: "/filter_mask"
+      base: 100.0
+      multiplier: -1.0
+  filter_mask_server:
+    ros__parameters:
+      use_sim_time: true
+      frame_id: "map"
+      topic_name: "/filter_mask"
```
SteveMacenski, line 233: Lets chose a less generic name than this for the topic, keep in mind this is the tutorial that people will use as a design reference so lets use a topic that's actually indicative of speed filter

tutorials/docs/navigation2_with_speed_filter.rst
```diff
@@ -0,0 +1,327 @@
...
+      yaml_filename: "speed_mask.yaml"
+
+Note, that:
+
+ - For Speed Filter setting speed restrictions in a percent from maximum speed, the ``type`` of costmap filter should be set to ``1``. All possible costmap filter types could be found at :ref:`configuring_map_server` page.
+ - Filter mask topic name should be the equal for ``mask_topic`` parameter of Costmap Filter Info Publisher Server and ``topic_name`` parameter of Map Server.
+ - As was described in a previous chapter, ``base`` and ``multiplier`` should be set to ``100.0`` and ``-1.0`` accordingly.
```
SteveMacenski, line 240:
```suggestion
 - As was described in a previous chapter, ``base`` and ``multiplier`` should be set to ``100.0`` and ``-1.0`` accordingly for the purposes of this tutorial example.
```

tutorials/docs/navigation2_with_speed_filter.rst
```diff
@@ -0,0 +1,327 @@
...
+
+Costmap Filters are Costamp2D plugins. You can enable the ``SpeedFilter`` plugin in Costmap2D by adding ``speed_filter`` to the ``plugins`` parameter in ``nav2_params.yaml``. The Speed Filter plugin should have the following parameters defined:
+
+- ``plugin``: type of plugin. In our case ``nav2_costmap_2d::SpeedFilter``.
+- ``filter_info_topic``: filter info topic name. This needs to be equal to ``filter_info_topic`` parameter of Costmap Filter Info Publisher Server from the chapter above.
+- ``speed_limit_topic``: name of topic to publish speed limit to.
+
+Full list of parameters supported by ``SpeedFilter`` are listed at the :ref:`speed_filter` page.
+
+You can place the plugin either in the ``global_costmap`` section in ``nav2_params.yaml`` to have speed restriction mask applied to global costmap or in the ``local_costmap`` to apply speed mask to the local costmap. However, it is highly not recommended to enable ``SpeedFilter`` plugin simultaneously for global and local costmaps. This might produce unwanted multiple "speed restriction" - "no restriction" message chains on speed restriction boundaries, that will cause jerking of the robot or another unpredictable behaviour.
```
SteveMacenski, line 265: Please be more stern than a recommendation, we do not want anyone to have them in both because it would really mess things up

tutorials/docs/navigation2_with_speed_filter.rst
```diff
@@ -0,0 +1,327 @@
...
+
+As stated in the `design <https://github.com/ros-planning/navigation2/blob/main/doc/design/CostmapFilters_design.pdf>`_, Speed Filter publishes speed restricting `messages <https://github.com/ros-planning/navigation2/blob/main/nav2_msgs/msg/SpeedLimit.msg>`_ targeted for a Controller Server so that it could restrict maximum speed of the robot when it needed. Controller Server has a ``speed_limit_topic`` ROS parameter for that, which should be set to the same as in ``speed_filter`` plugin value.
```
SteveMacenski, line 267:
```suggestion
As stated in the `design <https://github.com/ros-planning/navigation2/blob/main/doc/design/CostmapFilters_design.pdf>`_, Speed Filter publishes speed restricting `messages <https://github.com/ros-planning/navigation2/blob/main/nav2_msgs/msg/SpeedLimit.msg>`_ targeted for a Controller Server so that it could restrict maximum speed of the robot when it needed. Controller Server has a ``speed_limit_topic`` ROS parameter for that, which should be set to the same as in ``speed_filter`` plugin value. This topic in the map server could also be used to any number of other speed-restricted applications beyond the speed limiting zones, such as dynamically adjusting maximum speed by payload mass.
```

tutorials/docs/navigation2_with_speed_filter.rst
```diff
@@ -0,0 +1,327 @@
...
+        ...
+        plugins: ["static_layer", "obstacle_layer", "inflation_layer", "speed_filter"]
+        ...
+        speed_filter:
+          plugin: "nav2_costmap_2d::SpeedFilter"
+          enabled: True
+          filter_info_topic: "/costmap_filter_info"
+          speed_limit_topic: "/speed_limit"
+
+For using Speed Filter with local costmap, use the following configuration:
```
SteveMacenski, line 285: We should choose one or the other local / global costmap to recommend in the tutorial.

SteveMacenski (changes requested):

tutorials/docs/navigation2_with_speed_filter.rst
AlexeyMerzlyakov: Yes, that is because of in the example we want to have 1:1 mask grayscale % -> speed limit % conversion. From the tutorial:
> We will use scale map mode with no thresholds.

This mean, that `occupied_thresh = 1.0` and `free_thresh = 0.0`.

tutorials/docs/navigation2_with_speed_filter.rst
AlexeyMerzlyakov: Moved in https://github.com/ros-planning/navigation.ros.org/pull/124/commits/0ca276abe001d1fddab1d0caf89e8a9d15a95f5c#diff-8e354ab7d89ece0f0ce8893b7a0ac5fd4a8c11335c98fd9073499bb896162d95R93

tutorials/docs/navigation2_with_speed_filter.rst
AlexeyMerzlyakov: Updated to `/speed_filter_mask` in https://github.com/ros-planning/navigation.ros.org/commit/0ca276abe001d1fddab1d0caf89e8a9d15a95f5c#diff-8e354ab7d89ece0f0ce8893b7a0ac5fd4a8c11335c98fd9073499bb896162d95L226.
Plus updated in the same way for the Keepout tutorial: https://github.com/ros-planning/navigation.ros.org/commit/0ca276abe001d1fddab1d0caf89e8a9d15a95f5c#diff-9de6c0db49184c10da140edc0e8a0b610ce2186d7c72119697d77b581184d93cL208.

Also, updated the tutorials repo accordingly to fit the documentation:
https://github.com/ros-planning/navigation2_tutorials/pull/18/commits/284e4bed38804d582b479dc4b4a1fc008f519921

One thing was untouched - is a demo videos for Keepout and Speed filter. I do not think that to re-take whole video only because of parameter name change is a reasonable idea. However, correct me if I'm wrong.

tutorials/docs/navigation2_with_speed_filter.rst
AlexeyMerzlyakov: This was changed to ban starting from https://github.com/ros-planning/navigation.ros.org/pull/124/commits/0ca276abe001d1fddab1d0caf89e8a9d15a95f5c#diff-8e354ab7d89ece0f0ce8893b7a0ac5fd4a8c11335c98fd9073499bb896162d95L265

tutorials/docs/navigation2_with_speed_filter.rst
AlexeyMerzlyakov: Fixed stating from https://github.com/ros-planning/navigation.ros.org/commit/0ca276abe001d1fddab1d0caf89e8a9d15a95f5c#diff-8e354ab7d89ece0f0ce8893b7a0ac5fd4a8c11335c98fd9073499bb896162d95L269

Also, arranged some sentences there to make the narration more consistent:
https://github.com/ros-planning/navigation.ros.org/pull/124/commits/0ca276abe001d1fddab1d0caf89e8a9d15a95f5c#diff-8e354ab7d89ece0f0ce8893b7a0ac5fd4a8c11335c98fd9073499bb896162d95L267
->
https://github.com/ros-planning/navigation.ros.org/pull/124/commits/0ca276abe001d1fddab1d0caf89e8a9d15a95f5c#diff-8e354ab7d89ece0f0ce8893b7a0ac5fd4a8c11335c98fd9073499bb896162d95R279

tutorials/docs/navigation2_with_speed_filter.rst
SteveMacenski: That should be made really clear why you're doing this and that its important, I would have just scanned right through that without thinking that it didn't apply to this as much.

tutorials/docs/navigation2_with_speed_filter.rst
SteveMacenski: yeah, don't update a video for small details.

SteveMacenski (approved):
Just that one other item but otherwise no other review comments

tutorials/docs/navigation2_with_speed_filter.rst
AlexeyMerzlyakov: Added a sentence about thresholds choice in https://github.com/ros-planning/navigation.ros.org/pull/124/commits/26a9ceea8da06b9bdd8e5a00694e35ce46add883

tutorials/docs/navigation2_with_speed_filter.rst
```diff
@@ -0,0 +1,307 @@
...
+
+.. image:: images/Navigation2_with_Speed_Filter/drawing_speed_mask.png
+    :width: 500px
+
+Area "A" is filled with ``40%`` gray color, area "B" - with ``70%`` gray, that means that speed restriction will take ``100% - 40% = 60%`` in area "A" and ``100% - 70% = 30%`` in area "B" from maximum speed value allowed for this robot.
+We will use ``scale`` map mode with no thresholds. In this mode darker colors will have higher ``OccupancyGrid`` values. E.g. for area "B" with ``70%`` of gray ``OccupancyGrid`` data will be equal to ``70``. So in order to hit the target, we need to choose ``base = 100.0`` and ``multiplier = -1.0``. This will reverse the scale ``OccupancyGrid`` values to a desired one. No thresholds were chosen for the convenience: to have 1:1 full range conversion of lightness value from filter mask -> to speed restriction percent.
```
SteveMacenski, line 65:
```suggestion
We will use ``scale`` map mode with no thresholds. In this mode darker colors will have higher ``OccupancyGrid`` values. E.g. for area "B" with ``70%`` of gray ``OccupancyGrid`` data will be equal to ``70``. So in order to hit the target, we need to choose ``base = 100.0`` and ``multiplier = -1.0``. This will reverse the scale ``OccupancyGrid`` values to a desired one. No thresholds (``free_thresh`` ``occupied_thresh``) were chosen for the convenience in the ``yaml``  file: to have 1:1 full range conversion of lightness value from filter mask -> to speed restriction percent.
```

SteveMacenski (commented):
Make this one clarification and LGTM to merge

SteveMacenski (approved):

PR #131, opened 2021-01-13
Documentation changes for ray tracing param changes

danieljeswin:
## Description 

* Documenting the `raytrace_min_range` and `raytrace_max_range` params introduced in this [pull request](https://github.com/ros-planning/navigation2/pull/2126).

### DIFF
--- a/migration/Foxy.rst
+++ b/migration/Foxy.rst
@@ -113,7 +113,7 @@ To follow the SI units outlined in REP-103 to the "T" nodes below were modified 
 
 Ray Tracing Parameters
 **********************
-Raytracing functionality was modified to include a minimum range parameter from which ray tracing starts to clear obstacles to avoid incorrectly clearing obstacles too close to the robot. This issue was mentioned in `ROS Answers <https://answers.ros.org/question/355150/obstacles-in-sensor-deadzone/>`_. An existing parameter `raytrace_range` was renamed to `raytrace_max_range` to reflect the functionality it affects. The renamed parameters and the plugins that they belong to are mentioned below. The changes were introduced in this `pull request <https://github.com/ros-planning/navigation2/pull/2126>`_.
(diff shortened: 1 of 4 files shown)


migration/Foxy.rst
```diff
@@ -109,3 +109,15 @@ To follow the SI units outlined in REP-103 to the "T" nodes below were modified
  - ``max_planning_time_ms`` became ``max_planning_time`` in seconds
 - map saver
  - ``save_map_timeout`` in seconds
+
+
+Ray Tracing Parameters
+**********************
+Raytracing functionality was modified to include a minimum range parameter from which ray tracing starts to clear obstacles to avoid incorrectly clearing obstacles too close to the robot. This issue was mentioned in `ROS Answers <https://answers.ros.org/question/355150/obstacles-in-sensor-deadzone/>`_. An existing parameter `raytrace_range` was renamed to `raytrace_max_range` to reflect the functionality it affects. The renamed parameters and the plugins that they belong to are mentioned below. The changes were introduced in this `pull request <https://github.com/ros-planning/navigation2/pull/2126>`_.
```
SteveMacenski, line 116: All of your parameters need to be in double "`"'s for rst markdown you need 2 around code blocks to render - just a small thing

SteveMacenski (changes requested):
LGTM, just need to fix one thing that's not quite rendering

SteveMacenski:
by the way, if you hit "show all checks" and go to the job "details", you can get access to your PR's version of the website. Hit "artifacts" and "html/index.html" and it'll open up generated website you can surf to see how your changes look

SteveMacenski:
I just realized that the parameter is `raytrace_min_range` but we're also apply it to the marking not just raytrace clearing. 

(thinking aloud) we could change it to `obstacle_min_range` and `obstacle_max_range` and the min is also applied to raytracing. We could also make separate marking and clearing `min` variables. Finally, we could make a new name for this that's distinctly not `obstacle` or `raytrace` and we say its applied to both, though I think that's the worse of the  3 options.

What do you think? Sorry I didn't catch this sooner, this is actually why the docs can help alot, seeing it all next to each other gave me an "ah ha" moment. I think separate marking and clearing min variables would be the way to go. That wouldn't involve too much rework, just storing the second `min` variable and applying them in the `mark` functions.

danieljeswin:
Yeah I think having seperate marking and clearing variables is the way to go

SteveMacenski:
Awesome, sorry I didn't catch that sooner. Great work!

danieljeswin:
No worries! I have added a minimum range parameter for it and renamed the existing parameter as the max range parameter

SteveMacenski (approved):

PR #133, opened 2021-01-23
Update all instances of `laser_link` to `base_laser` for consistency

deepaktalwardt:
This PR fixes inconsistencies in TF frame names in the "Setting Up Transformations" guide. The frame name for the laser appears as `base_laser` in all figures, however, it is referred to as `laser_link` in many places within the document which can confuse readers. This PR changes all instances of `laser_link` to `base_laser` in this document.


SteveMacenski (approved):
Thanks for the detail update!

PR #152, opened 2021-03-22
planner_selector in migration guide

pabloinigoblasco:
Simple documentation to include the planner_selector on the migration guide.

### DIFF
--- a/configuration/packages/bt-plugins/actions/PlannerSelector.rst
+++ b/configuration/packages/bt-plugins/actions/PlannerSelector.rst
@@ -1,0 +1,56 @@
+.. _bt_planner_selector_node:
+
+PlannerSelector
+==============
+
+It is used to switch the planner that will be used by the planner server. It subscribes to the "planner_selector" topic to get name of the planner to be used. It is used before of the ComputePathToPoseAction. The selected_planner output port is passed to planner_id input port of the ComputePathToPoseAction.
+
+.. _bt_navigator: https://github.com/ros-planning/navigation2/tree/main/nav2_bt_navigator
+
+Input Ports
+-----------
+
+:topic_name:
+
+  ====== =======
+  Type   Default
+  ------ -------
+  string planner_selector  
+  ====== =======
+
+  Description
+    	The name of the topic used to received select command messages. This is used to support multiple PlannerSelector nodes. 
+      
+:default_planner:
+
+  ====== =======
+  Type   Default
+  ------ -------
+  string N/A  
+  ====== =======
+
+  Description
+    	The default value for the selected planner if no message is received from the input topic.
+
+
+Output Ports
+-----------
+
+:selected_planner:
+
+  ====== =======
+  Type   Default
+  ------ -------
+  string N/A  
+  ====== =======
+
+  Description
+    	The output selected planner id. This selected_planner string is usually passed to the ComputePathToPose behavior via the planner_id input port.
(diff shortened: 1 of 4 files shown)


plugins/index.rst
```diff
@@ -244,7 +244,8 @@ Behavior Tree Nodes
 +--------------------------------------------+---------------------+------------------------------------------+
 | `Truncate Path`_                           | Francisco Martín    | Modifies a path making it shorter        |
 +--------------------------------------------+---------------------+------------------------------------------+
-
+| `Planner Selector`_                        | Pablo Iñigo Blasco  | Selects the global planner               |
```
SteveMacenski, line 247: "based on a topic input, otherwises uses a default planner id"

migration/Foxy.rst
```diff
@@ -199,3 +199,10 @@ These plugins are set as default in the ``nav2_bt_navigator`` but may be overrid
 Original GitHub tickets:
 
 - `SingleTrigger <https://github.com/ros-planning/navigation2/pull/2236>`_
+
```
SteveMacenski, line 202: Just above this is the `SingleTrigger` added BT node section, you can remove this new section and just add this to the list after SingleTrigger with a link to the PR

configuration/packages/bt-plugins/actions/PlannerSelector.rst
```diff
@@ -0,0 +1,56 @@
+.. _bt_planner_selector_node:
```
SteveMacenski, line 1: There's no link to this page anywhere, please add this page to the list https://692-240847415-gh.circle-artifacts.com/0/html/configuration/packages/configuring-bt-xml.html on this page under Action Plugins to have this page then properly linked

configuration/packages/bt-plugins/actions/PlannerSelector.rst
```diff
@@ -0,0 +1,56 @@
...
+
+PlannerSelector
+==============
+
+It is used to select the planner that will be used by the planner server. It subscribes to the _planner_selector_ topic to recive command msgs with the name of the planner to be used. It is commonly used before of the ComputePathToPoseAction. The selected_planner output port is passed to planner_id input port of the ComputePathToPoseAction.
```
SteveMacenski, line 6:
```suggestion
It is used to select the planner that will be used by the planner server. It subscribes to the ``planner_selector`` topic to receive command messages with the name of the planner to be used. It is commonly used before of the ComputePathToPoseAction. The ``selected_planner`` output port is passed to ``planner_id`` input port of the ComputePathToPoseAction. If none is provided on the topic, the ``default_planner`` is used.
```

SteveMacenski (changes requested):

SteveMacenski:
https://github.com/ros-planning/navigation.ros.org/blob/master/configuration/packages/configuring-bt-xml.rst Last thing missing is your rst file has to be added to the Action list here so that it can be accessed

configuration/packages/bt-plugins/actions/PlannerSelector.rst
pabloinigoblasco: This should be already done:
https://github.com/ros-planning/navigation.ros.org/pull/152/commits/3bc2462254a2e33bb2e1ecc959b546d9275edc53

SteveMacenski (approved):

PR #157, opened 2021-03-31
Changed PointCloud msg publishers to PointCloud2 

saching13:
Added the details on where PointCloud msg Publishers and data is changed to use PointCloud2 msg in Fory to Galactic migration.


SteveMacenski (approved):

mikeferguson:
@SteveMacenski does this actually belong in a Galactic Migration document, not Foxy migration? This is a new feature in Galactic, right?

PR #160, opened 2021-04-15
Fix default failure_tolerance

doisyg:
to be consistent with https://github.com/ros-planning/navigation2/blob/f28951048dc67f1e8a7b1b7cf450195e53a80ee6/nav2_bringup/bringup/params/nav2_params.yaml#L106


SteveMacenski:
https://github.com/ros-planning/navigation2/blob/81e1c7d0722af03e9aa4f82eeac5b0f0bcd6fbf7/nav2_controller/src/nav2_controller.cpp#L59

The default is set in the code, the values in the yaml file are not reflective of the defaults if no value was set

doisyg:
Oups, sorry (and I am the one who wrote it...)

PR #161, opened 2021-04-21
Update for costmap plugins and filters separation

AlexeyMerzlyakov:
This is documentation update related to https://github.com/ros-planning/navigation2/pull/2308 PR.


configuration/packages/configuring-costmaps.rst
```diff
@@ -295,6 +295,33 @@ Costmap2D ROS Parameters
...
+  Type           Default
+  -------------- -------
+  vector<string> {}
+  ============== =======
+
+  Description
+    List of mapped costmap filter names for parameter namespaces and names.
+
+  Note
+    Actually, costmap filters are also loadable plugins. The separation is made to avoid plugins and filters work interferences. As well as for common plugins, each costmap filter namespace defined in this list needs to have a :code:`plugin` parameter defining the type of filter plugin to be loaded in the namespace.
```
SteveMacenski, line 310:
```suggestion
    Costmap filters are also loadable plugins just as ordinary costmap layers. This separation is made to avoid plugin and filter interference and places these filters on top of the combined layered costmap. As  with plugins, each costmap filter namespace defined in this list needs to have a :code:`plugin` parameter defining the type of filter plugin to be loaded in the namespace.
```

tutorials/docs/navigation2_with_keepout_filter.rst
```diff
@@ -273,7 +275,7 @@ To enable ``KeepoutFilter`` with same mask for both global and local costmaps, u
 
 .. note::
 
-  In this example ``keepout_filter`` goes after the ``inflation_layer`` so the keepout boundaries aren't inflated as obstacles, although this is not an obligatory requirement.
+  All costmap filters should be enabled through a ``filters`` parameter. This is made for separation of plugins and filters work. For keepout filter this prevents the keepout boundaries from inflation as obstacles.
```
SteveMacenski, line 278:
```suggestion
  All costmap filters should be enabled through a ``filters`` parameter -- though it is technically possible to include in the layered costmap itself. This is separated from the layer plugins to prevent interference in the layers, particularly the  inflation layer.
```

SteveMacenski (changes requested):

PR #162, opened 2021-04-23
Add Odometry Setup Guide

jd-deleon:
Co-authored-by: Areeya Rubenecia <a.rubenecia@samsung.com>
Co-authored-by: Fiona Estoperez <f.estoperez@samsung.com>
Co-authored-by: Steve Macenski <stevenmacenski@gmail.com>


SteveMacenski:
Does this also need an update (including files) to the sambot description in the tutorials package?

SteveMacenski (approved):
LGTM! Just need a PR to update the files in the tutorials package to match and we can merge this in!

jd-deleon:
> Does this also need an update (including files) to the sambot description in the tutorials package?

Yes, just wanted to check first if  we are already good here before we file the PR to the tutorials repo. Do you think it will be better if we create a new package (sam_bot_odometry) in the tutorials repository that is separate from the description package? This way, our readers will be able to view the different stages of the robot setup as they go through the guide.

SteveMacenski:
I think all in 1 is OK

jd-deleon:
Done. Updated images, information and nav2 tutorials repository in latest commits

PR #182, opened 2021-06-18
update the params for nav2_theta_star_planner

Anshu-man567:


plugins/index.rst
```diff
@@ -96,6 +96,36 @@ Controllers
 Planners
 ========
 
+<<<<<<< HEAD
```
SteveMacenski, line 99: err umm I think you didn't fix a merge conflict very well :-)

SteveMacenski (changes requested):

plugins/index.rst
Anshu-man567: give me a minute :sweat_smile:

PR #187, opened 2021-07-07
Update links to ROS 2 docs

christophebedard:
docs.ros.org has replaced index.ros.org/doc/ros2

I used `/rolling/` for the links because then they won't need to be updated.

Signed-off-by: Christophe Bedard <bedard.christophe@gmail.com>


SteveMacenski (approved):

SteveMacenski:
Thanks!

PR #194, opened 2021-08-03
Add Sensors Setup Guide

jd-deleon:
Co-authored-by: Areeya Rubenecia a.rubenecia@samsung.com
Co-authored-by: Fiona Estoperez f.estoperez@samsung.com
Co-authored-by: Steve Macenski stevenmacenski@gmail.com


setup_guides/sensors/setup_sensors.rst
```diff
@@ -0,0 +1,569 @@
...
+In this subsection, we discuss some of the common types of ``sensor_msgs`` you might encounter when setting up Nav2. We will provide a brief description for each sensor, an image of it being simulated in Gazebo and the corresponding visualization of the sensor readings in RViz.
+
+.. note::  There are other types of ``sensor_msgs`` aside from the ones listed below.  The complete list of messages and their definitions can be found in the `sensor_msgs documentation <http://wiki.ros.org/sensor_msgs>`_.
+
+sensor_msgs/LaserScan
+---------------------
+
+This message represents a single scan from a planar laser range-finder. This message is used in ``slam_toolbox`` and ``nav2_amcl`` for localization and mapping, or in ``nav2_costmap_2d`` for perception.
+
+.. image:: images/sensor_laserscan_gazebo.png
```
SteveMacenski, line 33: For all image pairs, can you get them to render in the middle of the page? These 2 are nearly 50% so they effectively so, but the next one for instance is off to the left noticeably.

setup_guides/sensors/setup_sensors.rst
```diff
@@ -0,0 +1,569 @@
...
+
+.. image:: images/sensor_image_gazebo.png
+    :width: 43 %
+
+.. image:: images/sensor_image_rviz.png
+    :width: 49 %   
+
+Simulating Sensors using Gazebo
+*******************************
+To give you a better grasp of how to set up sensors on a simulated robot, we will build up on our previous tutorials and attach sensors to our simulated robot ``sam_bot``. Similar to the previous tutorial where we used Gazebo plugins to add odometry sensors to ``sam_bot``, we will be using the Gazebo plugins to simulate a lidar sensor and a depth camera on ``sam_bot``.
```
SteveMacenski, line 76:
```suggestion
To give you a better grasp of how to set up sensors on a simulated robot, we will build up on our previous tutorials and attach sensors to our simulated robot ``sam_bot``. Similar to the previous tutorial where we used Gazebo plugins to add odometry sensors to ``sam_bot``, we will be using the Gazebo plugins to simulate a lidar sensor and a depth camera on ``sam_bot``. If you are working with a real robot, most of these steps are still required for setting up your URDF frames and it will not hurt to also add in the gazebo plugins for later use. 
```

setup_guides/sensors/setup_sensors.rst
```diff
@@ -0,0 +1,569 @@
...
+
+You should also see the ``sensor_msgs/PointCloud2``, as shown below.
+
+.. image:: images/demo_pointcloud2_rviz.png
+    :align: center
+
+
+Mapping and Localization
+************************
+Now that we have a robot with its sensors set up, we can use the obtained sensor information to build a map of the environment and to localize the robot on the map. The ``slam_toolbox`` package is a set of tools and capabilities for 2D Simultaneous Localization and Mapping (SLAM) in potentially massive maps with ROS2. It is also one of the officially supported SLAM library by Nav2, and we recommend to use this package in situations you need to use SLAM on your robot setup. Aside from the ``slam_toolbox``, localization can also be implemented through the ``nav2_amcl`` package. This package implements Adaptive Monte Carlo Localization (AMCL) which estimates the position and orientation of the robot in a map. Other techniques may also be available, please check Nav2 documentation for more information.
```
SteveMacenski, line 319:
```suggestion
Now that we have a robot with its sensors set up, we can use the obtained sensor information to build a map of the environment and to localize the robot on the map. The ``slam_toolbox`` package is a set of tools and capabilities for 2D Simultaneous Localization and Mapping (SLAM) in potentially massive maps with ROS2. It is also one of the officially supported SLAM libraries in Nav2, and we recommend to use this package in situations you need to use SLAM on your robot setup. Aside from the ``slam_toolbox``, localization can also be implemented through the ``nav2_amcl`` package. This package implements Adaptive Monte Carlo Localization (AMCL) which estimates the position and orientation of the robot in a map. Other techniques may also be available, please check Nav2 documentation for more information.
```

setup_guides/sensors/setup_sensors.rst
```diff
@@ -0,0 +1,569 @@
...
+
+The costmap implementation consists of multiple layers, each of which has a certain function that contributes to a cell's overall cost. It consists of the following layers: static layer, inflation layer, range layer, obstacle layer, and voxel layer. The static layer represents the constant section of the costmap, like those produced by SLAM. The obstacle layer includes the obstacles reported by sensors in 2D. The voxel layer is similar to the obstacle layer but handles 3D data instead. The range layer allows for the inclusion of sonar and infrared data to our costmap. Lastly, the inflation layer represents the added cost values around lethal obstacles such that our robot avoids navigating close to obstacles. In the next subsection of this tutorial, we will have some discussion about the basic configuation of the different layers in ``nav2_costmap_2d``.
```
SteveMacenski, line 335:
```suggestion
The costmap implementation consists of multiple layers, each of which has a certain function that contributes to a cell's overall cost. The package consists of the following layers, but are plugin-based to allow customization and new layers to be used as well: static layer, inflation layer, range layer, obstacle layer, and voxel layer. The static layer represents the map section of the costmap, like those produced by SLAM. The obstacle layer includes the obstacles reported by sensors in 2D. The voxel layer is similar to the obstacle layer but handles 3D data instead. The range layer allows for the inclusion of sonar and infrared data to our costmap. Lastly, the inflation layer represents the added cost values around lethal obstacles such that our robot avoids navigating into obstacles due to the robot's geometry. In the next subsection of this tutorial, we will have some discussion about the basic configuration of the different layers in ``nav2_costmap_2d``. 
```

setup_guides/sensors/setup_sensors.rst
```diff
@@ -0,0 +1,569 @@
...
+
+The layers are integrated into the costmap through a plugin interface and then inflated using a user-specified `inflation radius <http://wiki.ros.org/costmap_2d/hydro/inflation>`_, if the inflation layer is enabled. For a deeper discussion on costmap concepts, you can have a look at the `ROS1 costmap_2D documentation <http://wiki.ros.org/costmap_2d>`_. Note that the ``nav2_costmap_2d`` package is mostly a straightforward ROS2 port of the ROS1 navigation stack version with minor changes required for ROS2 support and some new layer plugins.
+
+Configuring Costmap_2D
+======================
+
+In this subsection, we will show an example configuration of ``costmap_2d`` and explain the basic configuration parameters. The code block below is an example of ``costmap_2d`` configuration using static layer, obstacle layer, voxel layer, and inflation layer. Note that this block of parameters will be included in the `config/nav2_params.yaml <https://github.com/ros-planning/navigation2_tutorials/blob/master/sam_bot_description/config/nav2_params.yaml>`_ file  when we launch Nav2 at the end of this section.
```
SteveMacenski, line 342:
```suggestion
In this subsection, we will show an example configuration of ``costmap_2d`` and explain the basic configuration parameters for your sensors. The code block below is an example of ``costmap_2d`` configuration using static layer, obstacle layer, voxel layer, and inflation layer. Note that this block of parameters will be included in the `config/nav2_params.yaml <https://github.com/ros-planning/navigation2_tutorials/blob/master/sam_bot_description/config/nav2_params.yaml>`_ file  when we launch Nav2 at the end of this section.
```

setup_guides/sensors/setup_sensors.rst
```diff
@@ -0,0 +1,569 @@
...
+      ros__parameters:
+        update_frequency: 1.0
+        publish_frequency: 1.0
+        global_frame: map
+        robot_base_frame: base_link
+        use_sim_time: True
+        robot_radius: 0.3
+        resolution: 0.05
+        track_unknown_space: true
+        rolling_window: true
```
SteveMacenski, line 358: This should be false for most situations, its only true generally for the local costmap

setup_guides/sensors/setup_sensors.rst
```diff
@@ -0,0 +1,569 @@
...
+Launching slam_toolbox
+----------------------
+
+To be able to launch ``slam_toolbox``, make sure that you have installed the ``slam_toolbox`` package by executing the following command:
+
+.. code-block:: shell
+
+  sudo apt install ros-<ros2-distro>-slam-toolbox
+
+Now, we will configure the parameters of ``slam_toolbox``. Go to the ``config`` directory and create a ``mapper_params.yaml`` file. Then copy the contents of `config/mapper_params.yaml <https://github.com/ros-planning/navigation2_tutorials/blob/master/sam_bot_description/config/mapper_params.yaml>`_ and paste them into ``mapper_params.yaml``.
```
SteveMacenski, line 470: So I think here just use my parameter files / launch files in SLAM Toolbox itself, not adding a new launch/config file in this package here (especially just for a demo unless you've remapped the scan topics to non-default `scan`). If you have, I think you should remove the `demo/` namespace since that isn't standard

setup_guides/sensors/setup_sensors.rst
```diff
@@ -0,0 +1,569 @@
...
+
+Now, we will configure the parameters of ``slam_toolbox``. Go to the ``config`` directory and create a ``mapper_params.yaml`` file. Then copy the contents of `config/mapper_params.yaml <https://github.com/ros-planning/navigation2_tutorials/blob/master/sam_bot_description/config/mapper_params.yaml>`_ and paste them into ``mapper_params.yaml``. 
+
+The important thing to note in the configuration of ``mapper_params.yaml`` is that we make ``slam_toolbox`` use the laser data published in ``/demo/scan`` by setting the ``scan_topic`` parameter to ``/demo/scan``. Again, the in-depth discussion of other parameters is not included in this tutorial, please refer to the official documentation of `slam_toolbox <https://github.com/SteveMacenski/slam_toolbox#readme>`_  for more information on other parameters. 
+
+Before we launch the ``slam_toolbox``, make sure you have launched sam_bot's ``display.launch.py`` by executing the the lines below. Remember that this provides our ``odom`` => ``base_link`` and ``base_link`` => ``sensors`` transformations. It also launches RViz which we can use to visualize the robot and Gazebo which acts as our physics simulator.
```
SteveMacenski, line 474: The URDF doesn't provide the odom->baselink transformation. Gazebo does through the diff drive plugin

setup_guides/sensors/setup_sensors.rst
```diff
@@ -0,0 +1,569 @@
...
+Next, we will configure the costmap parameters of Nav2 before we launch it.
+
+First, Make sure you have installed the Nav2 packages by executing the following:
+
+.. code-block:: shell
+
+  sudo apt install ros-<ros2-distro>-navigation2
+  sudo apt install ros-<ros2-distro>-nav2-bringup
+
+We will configure the parameters of Nav2 to use our costmap configuration in the previous subsection. Create a file named ``nav2_params.yaml`` under the ``config`` directory and then copy the contents of `config/nav2_params.yaml <https://github.com/ros-planning/navigation2_tutorials/blob/master/sam_bot_description/config/nav2_params.yaml>`_ and paste them into the newly created ``nav2_params.yaml`` file.
```
SteveMacenski, line 518: As well here, if you remove the `demo` namespace, then you can just use the default parameters and launch files in Nav2 itself

setup_guides/sensors/setup_sensors.rst
```diff
@@ -0,0 +1,569 @@
...
+
+  colcon build
+  . install/setup.bash
+  ros2 launch sam_bot_description display.launch.py
+
+Then, we will launch the ``async_slam_toolbox_node`` of ``slam_toolbox`` and set its parameters to ``mapper_params.yaml`` using the package's built-in launch files. Open a new terminal and then execute the following lines:
+
+.. code-block:: shell
+
+  ros2 launch slam_toolbox online_async_launch.py slam_params_file:=<full/path/to/config/mapper_params.yaml>
```
SteveMacenski, line 486: Where is gazebo launched?

setup_guides/sensors/setup_sensors.rst
SteveMacenski: Also, where is gazebo launched in this instruction?

SteveMacenski (changes requested):
The costmap_2d section needs a bit of work, but overall great! I think it just needs to be simplified so we can leverage the default configurations/launch files in the packages by removing the gazebo `demo/` namespace to our data sources. 

Also, I think an edit over this section remembering that our goal is to explain how to set up our sensors w.r.t. costmap_2d might be good. There are a couple of times where it goes in to explaining more info as an overview of costmap_2d rather than focusing from the perspective of 'setting up my sensors for the first time'. I don't think any of the content is wrong, I think some of the sentences could be better tied back to the goal of the tutorial more directly

setup_guides/sensors/setup_sensors.rst
arubenecia: We already removed the demo namespace and removed the new config file for slam toolbox.

setup_guides/sensors/setup_sensors.rst
arubenecia: The display.launch.py also launches Gazebo so the odom->baselink transform should be provided upon launching display.launch.py. We will make clarifications about this on the draft.

setup_guides/sensors/setup_sensors.rst
arubenecia: The demo namespace and the new config file for nav2 have also been removed

setup_guides/sensors/setup_sensors.rst
arubenecia: Gazebo is launched inside the display.launch.py

setup_guides/sensors/setup_sensors.rst
SteveMacenski: They are still included in the tutorials PR

setup_guides/sensors/setup_sensors.rst
SteveMacenski: OK, make sure to update the tutorials PR

arubenecia:
The tutorial is now updated. Changes made according to the comments are included in the latest commits.

setup_guides/sensors/setup_sensors.rst
```diff
@@ -433,17 +409,17 @@ Note that the obstacle layer and voxel layer can use either or both ``LaserScan`
...
+For the voxel layer (lines 51-66), we set the ``publish_voxel_map`` parameter to ``True`` to enable the publishing of the 3D voxel grid. The resolution of the voxels in height is defined using the ``z_resolution`` parameter, while the number of voxels in each column is defined using the ``z_voxels`` parameter. The ``mark_threshold`` parameter sets the minimum number of voxels in a column to mark as occupied in the occupancy grid. We set the ``observation_sources`` parameter of the voxel layer to ``scan``, and we set the scan parameters (in lines 61-66) similar to the parameters that we have discussed for the obstacle layer. As defined in its ``topic`` and ``data_type`` parameters, the voxel layer will use the ``LaserSCan`` published on the ``/scan`` topic by the lidar scanner.
```
SteveMacenski, line 422:
```suggestion
For the voxel layer (lines 51-66), we set the ``publish_voxel_map`` parameter to ``True`` to enable the publishing of the 3D voxel grid. The resolution of the voxels in height is defined using the ``z_resolution`` parameter, while the number of voxels in each column is defined using the ``z_voxels`` parameter. The ``mark_threshold`` parameter sets the minimum number of voxels in a column to mark as occupied in the occupancy grid. We set the ``observation_sources`` parameter of the voxel layer to ``scan``, and we set the scan parameters (in lines 61-66) similar to the parameters that we have discussed for the obstacle layer. As defined in its ``topic`` and ``data_type`` parameters, the voxel layer will use the ``LaserScan`` published on the ``/scan`` topic by the lidar scanner.
```

SteveMacenski (changes requested):

SteveMacenski (approved):

Issue #250, opened 2021-12-14
pure pursuit use_velocity_scaled_lookahead_dist


joe28965:
On the website on [this page](https://navigation.ros.org/configuration/packages/configuring-regulated-pp.html) the `use_velocity_scaled_lookahead_dist` is said to have the default `double` of `0.4`. That should just be a `bool` set to `false` right?
I think it's this [github page](https://github.com/ros-planning/navigation.ros.org/blob/master/configuration/packages/configuring-regulated-pp.rst) that should be edited to update the `use_velocity_scaled_lookahead_dist` right?

I'm not 100% sure if the default is set to `false` though.

SteveMacenski:
That is correct- can you submit a PR to fix?

joe28965:
done. PR #251

PR #251, opened 2021-12-15
changed use_velocity_scaled_lookahead_dist to correct default (#250)

joe28965:


SteveMacenski (approved):
Much thanks!! Everyone doing their part helps keep the documentation improving. Sometimes I drone on when writing it and that can cause errors.

PR #266, opened 2022-01-28
Smac admissible heuristic update

afrixs:
Added notes about smac planner heuristic admissibility fix and retrospective penalty

### DIFF
--- a/configuration/packages/smac/configuring-smac-hybrid.rst
+++ b/configuration/packages/smac/configuring-smac-hybrid.rst
@@ -167,6 +167,17 @@ Parameters
   Description
     Heuristic penalty to apply to SE2 node for cost at pose. Allows Hybrid-A* to be cost aware.
 
+:``<name>``.retrospective_penalty:
+
+  ====== =======
+  Type   Default                                                   
+  ------ -------
+  double 0.025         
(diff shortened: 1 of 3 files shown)


configuration/packages/smac/configuring-smac-hybrid.rst
```diff
@@ -167,6 +167,17 @@ Parameters
...
+:``<name>``.retrospective_penalty:
+
+  ====== =======
+  Type   Default                                                   
+  ------ -------
+  double 0.025         
+  ====== =======
+
+  Description
+    Heuristic penalty to apply to SE2 node penalty. Causes Hybrid-A* to prefer later maneuvers before earlier ones along the path. Saves search time since earlier (shorter) branches are not expanded until it is necessary. Must be >= 0.0 and <= 1.0.
```
SteveMacenski, line 179:
```suggestion
    Heuristic penalty to apply to SE2 node penalty. Causes Hybrid-A* to prefer later maneuvers before earlier ones along the path. Saves search time since earlier (shorter) branches are not expanded until it is necessary. Must be >= 0.0 and <= 1.0. Must be `0.0` to be fully admissible. 
```

PR #270, opened 2022-02-03
Fixed example for AMCL initial pose settings

moooeeeep:
The previous example didn't really translate to the yaml format that has to be filled in there.

### DIFF
--- a/configuration/packages/configuring-amcl.rst
+++ b/configuration/packages/configuring-amcl.rst
@@ -201,11 +201,11 @@ Parameters
 
 :initial_pose:
 
-  ============== =============================
-  Type           Default                                               
-  -------------- -----------------------------
-  Pose2D         {0.0, 0.0, 0.0, 0.0}         
-  ============== =============================
+  ============== ==================================
+  Type           Default                           
+  -------------- ----------------------------------
+  Pose2D         {x: 0.0, y: 0.0, z: 0.0, yaw: 0.0}
+  ============== ==================================
 
   Description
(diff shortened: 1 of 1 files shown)


moooeeeep:
Looking at other doc entries, it seems the format given in those tables is supposed to resemble C++ list-initialization instead of actual yaml format. So probably I misunderstood the format of these entries.

I feel giving yaml snippets there would be more useful, but others might disagree. Probably one should adjust this consistently (if at all), which this PR obviously doesn't attempt. Feel free to discuss/close.

configuration/packages/configuring-amcl.rst
```diff
@@ -204,7 +204,7 @@ Parameters
   ============== =============================
   Type           Default                                               
   -------------- -----------------------------
-  Pose2D         {0.0, 0.0, 0.0, 0.0}         
+  Pose2D         {x: 0.0, y: 0.0, z: 0.0, yaw: 0.0}
```
SteveMacenski, line 207: You need to extend the `===` and `---` lines so that the formatting of the table renders properly. See other examples in the file essentially the lengths of the fields need to match the structure of the table.

SteveMacenski (changes requested):

configuration/packages/configuring-amcl.rst
moooeeeep: No problem. I adjusted the markup accordingly.

SteveMacenski (approved):
Much appreciated, thanks!!

PR #276, opened 2022-02-10
Remove an extra parameter description in PreferForwardCritic page.

stu00608:
There are 2 "scale" descriptions in this page, and it seems like an extra content.


SteveMacenski (approved):
thanks!

PR #278, opened 2022-02-14
Clarification to `robot_model_type` for users of galactic and earlier

moooeeeep:
Added note about breaking change in AMCL motion model selection parameter.

### DIFF
--- a/configuration/packages/configuring-amcl.rst
+++ b/configuration/packages/configuring-amcl.rst
@@ -322,7 +322,7 @@ Parameters
     The fully-qualified type of the plugin class. Options are "nav2_amcl::DifferentialMotionModel" and "nav2_amcl::OmniMotionModel". Users can also provide their own custom motion model plugin type.
 
   Note for users of galactic and earlier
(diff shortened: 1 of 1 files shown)


configuration/packages/configuring-amcl.rst
```diff
@@ -319,7 +319,10 @@ Parameters
   ============== =============================
 
   Description
-    The fully-qualified type of the plugin class. Options are nav2_amcl::DifferentialMotionModel and nav2_amcl::OmniMotionModel.
+    The fully-qualified type of the plugin class. Options are "nav2_amcl::DifferentialMotionModel" and "nav2_amcl::OmniMotionModel". Users can also provide their own custom motion model plugin type.
+
+  Note for users of galactic and earlier
+    The above options are only available for users of a source build or the rolling release. Up until the galactic release only two models are selectable by string key (valid options: "differential", "omnidirectional").
```
SteveMacenski, line 325:
```suggestion
    The models are selectable by string key (valid options: "differential", "omnidirectional") rather than plugins.
```

PR #282, opened 2022-03-01
Added docs

tonynajjar:
Added docs for https://github.com/ros-planning/navigation2/pull/2834


tonynajjar:
See https://github.com/ros-planning/navigation.ros.org/pull/283

Issue #281, opened 2022-02-28
AMCL: Missing first_map_only_ parameter in docs


hardesh:
Hi,

`first_map_only_` is missing from the parameters. But it's being used in AMCL https://github.com/ros-planning/navigation2/blob/0a961df9cbe50356484877681abfdbeae6ba7b33/nav2_amcl/src/amcl_node.cpp#L1058 

It's being set to True by default in the header file. If we change the map using `/map_server/load_map` service then it should get reflected in AMCL as well i think

SteveMacenski:
Would you mind submitting a PR to add it? Thanks!

hardesh:
yeah sure

PR #284, opened 2022-03-02
Added first_map_only parameter in docs

hardesh:
Adresses #281

### DIFF
--- a/configuration/packages/configuring-amcl.rst
+++ b/configuration/packages/configuring-amcl.rst
@@ -467,6 +467,18 @@ Parameters
   Description
     Map topic to subscribe to.
 
+:first_map_only_:
+
+  ============== =============================
+  Type           Default
+  -------------- -----------------------------
+  bool           False
+  ============== =============================
+
+  Description
+    Allows AMCL to accept maps more than once on the map_topic. This is especially useful when you're using the `LoadMap` service in `map_server`.
+
+
 Example
 *******
 .. code-block:: yaml
@@ -513,6 +525,7 @@ Example
         map_topic: map
         set_initial_pose: false
         always_reset_initial_pose: false
+        first_map_only_: false
         initial_pose:
           x: 0.0
           y: 0.0
--- a/configuration/packages/configuring-amcl.rst
+++ b/configuration/packages/configuring-amcl.rst
@@ -467,7 +467,7 @@ Parameters
   Description
     Map topic to subscribe to.
 
-:first_map_only_:
+:first_map_only:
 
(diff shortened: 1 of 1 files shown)


configuration/packages/configuring-amcl.rst
```diff
@@ -467,6 +467,18 @@ Parameters
   Description
     Map topic to subscribe to.
 
+:first_map_only_:
```
SteveMacenski, line 470: No underscore, then LGTM!

configuration/packages/configuring-amcl.rst
```diff
@@ -513,6 +525,7 @@ Example
         map_topic: map
         set_initial_pose: false
         always_reset_initial_pose: false
+        first_map_only_: false
```
SteveMacenski, line 528: Remove the underscore here too!

SteveMacenski (changes requested):

configuration/packages/configuring-amcl.rst
hardesh: Hi Steve,
I think we'll have to change it in the code as well https://github.com/ros-planning/navigation2/blob/0a961df9cbe50356484877681abfdbeae6ba7b33/nav2_amcl/src/amcl_node.cpp#L1058

hardesh: only this parameter has underscore after it

SteveMacenski: https://github.com/ros-planning/navigation2/blob/6522fed52926f240736d628d28c11a0be8f238cb/nav2_amcl/src/amcl_node.cpp#L1058

oof, yeah that would be good to correct, thanks for bringing that up. Yes, we should remove the underscore here. That was clearly a copy-paste error when it was being ported from ROS 1. Thanks for mentioning that!

SteveMacenski:
Any update?

hardesh:
Should I also open a PR in nav2? I thought you were going to take this up :grimacing: 

Also if I fix this it'll be merged in `main` and it can't be backported to foxy right?

hardesh:
I can open a PR in nav2 tomorrow

SteveMacenski:
Yeah, opening a PR would be great! 

And no, unfortunately it can’t be backported since it would change behavior suddenly

SteveMacenski:
@hardesh Thanks for the PRs and pointing out this issue, it is much appreciated! :small_airplane:

hardesh:
No problem! :smiley:

PR #290, opened 2022-03-18
removing the annoying warnings after build

padhupradheep:
Cleaning up annoying warnings after you build, but there is still more warnings which needs time and I will address them later in a separate PR.


SteveMacenski (approved):
Thanks! Agreed, it just never quite bothered me enough to fix!

PR #308, opened 2022-05-05
Constrained smoother and TruncatePathLocal BT Node

afrixs:


SteveMacenski:
2 changes missing:

Need to add the BT node and the Smoother algorithm to the tables in the navigation algorithms page: https://github.com/ros-planning/navigation.ros.org/blob/master/plugins/index.rst. For the BT node, its all you, but for the smoother add both of our names since it was based on my initial work in Smac. 

Also need migration guide update to mention that hey this exists with a short description and a link to the configuration / PR.

configuration/packages/bt-plugins/actions/TruncatePathLocal.rst
```diff
@@ -0,0 +1,118 @@
...
+  ====== ===========
+  Type   Default
+  ------ -----------
+  string "base_link"
+  ====== ===========
+
+  Description
+    	Robot base frame id.
+
+:transform_tolerance:
```
SteveMacenski, line 55: Transform tolerance appears twice

configuration/packages/bt-plugins/actions/TruncatePathLocal.rst
```diff
@@ -0,0 +1,118 @@
...
+  double 0.2
+  ====== ===========
+
+  Description
+    	Robot pose lookup tolerance.
+
+Ouput Ports
+-----------
+
+:max_robot_pose_search_dist:
```
SteveMacenski, line 102: I think this is an input port?

Also, I think this is missing the output port for the `output_path`

configuration/packages/configuring-constrained-smoother.rst
```diff
@@ -0,0 +1,287 @@
...
+
+A smoother plugin for nav2_smoother based on the original deprecated smoother in nav2_smac_planner and put into operational state by `RoboTech Vision`_.
+Suitable for applications which need planned global path to be pushed away from obstacles and/or for Reeds-Shepp motion models.
+
+.. _`TruncatePathLocal BT Node`: bt-plugins/actions/TruncatePathLocal.html
+
+Important note: Constrained smoother uses a rather heavy optimization algorithm and thus is suggested to be used on a periodically truncated path.
+`TruncatePathLocal BT Node`_ can be used for achieving a proper path length and DistanceController BT Node can be used for achieving periodicity.
+
+.. image:: images/constrained_smoother/constrained_smoother.png
```
SteveMacenski, line 20: Make a note that the blue is original, green is smoothed (what's red?)

Also, make a note that the blue is a particularly bad example of a path to highlight the power of the smoother (that's not a usually expected output of Smac in its current form in Nav2)

configuration/packages/configuring-constrained-smoother.rst
```diff
@@ -0,0 +1,287 @@
...
+
+  Description
+    Minimum turning radius the robot can perform. Can be set to 0.0 (or w_curve can be set to 0.0 with the same effect) for diff-drive/holonomic robots
+
+:w_curve:
+
+  ============== ===========================
+  Type           Default                    
+  -------------- ---------------------------
+  double         30.0
```
SteveMacenski, line 99: Are these weights all tuned and reasonable default values? Just checking, we care alot about out-of-the-box quality in Nav2

SteveMacenski (changes requested):
Love it, really only a couple of changes but otherwise LGTM

configuration/packages/configuring-constrained-smoother.rst
afrixs: Yes, I think the smoother gives pretty good results with these values.

configuration/packages/configuring-constrained-smoother.rst
afrixs: The red is irrelevant here (it's published by our internal controller). Frankly, I just found this image among some old screenshots, didn't spend time creating a clearer one without extra visualizations... But I considered it good enough as an example, that's why I put it there...

Added the notes

SteveMacenski:
Changes are approved but still missing the 2 items from the comment above; 


```
Need to add the BT node and the Smoother algorithm to the tables in the navigation algorithms page: https://github.com/ros-planning/navigation.ros.org/blob/master/plugins/index.rst. For the BT node, its all you, but for the smoother add both of our names since it was based on my initial work in Smac.

Also need migration guide update to mention that hey this exists with a short description and a link to the configuration / PR.
```

PR #310, opened 2022-05-08
Update launch file instructions

omerts:
Current instruction miss a very important step, Without it the joint_state_publisher doesn't work, as the default param for `gui` is true. This adds the needed instructions for this tutorial to actually work.


SteveMacenski (approved):
Thanks for the contribution, I agree!

PR #313, opened 2022-05-21
Update writing_new_behavior_plugin.rst

Ryujiyasu:
1. fix parameters

2. parameter structure change now so i mention galactic and earlier

https://github.com/ros-planning/navigation2/pull/2867


Ryujiyasu:
Move to here

https://github.com/ros-planning/navigation.ros.org/pull/316

Issue #319, opened 2022-06-02
Not able to build on 22.04


padhupradheep:
Something is fishy, when I try to build on ubuntu 22.04

```
WARNING: Invalid configuration value found: 'language = None'. Update your configuration to a valid langauge code. Falling back to 'en' (English).

Exception occurred:
  File "/home/pradheep/.local/lib/python3.10/site-packages/breathe/node_factory.py", line 5, in <module>
    import sphinx.ext.mathbase
ModuleNotFoundError: No module named 'sphinx.ext.mathbase'
The full traceback has been saved in /tmp/sphinx-err-v_0nnovb.log, if you want to report the issue to the developers.
```

I'll check for the issue and will give a PR..

SteveMacenski:
Did you not install all necessary dependencies? It looks like `mathbase` just needs to be installed?

padhupradheep:
but it should be installed as part of the sphinx core right?

That is what the documentation says: https://www.sphinx-doc.org/en/master/usage/extensions/math.html

padhupradheep:
So, the solution would be to use breathe 4.28.0 ! I'll give a PR after testing it on 20.04 as well

PR #323, opened 2022-06-09
upgrading breathe version

padhupradheep:
works on 20.04 and 22.04 as well.. #319 

apt-get installation of sphinx installs sphinx 1.8.0.. 

The newer version of Breathe supports versions between 3.0.0 and 3.5.0. Therefore, adding a pip3 install targeting the supported versioning for sphinx.


SteveMacenski (approved):
Thanks!

PR #331, opened 2022-06-27
Update link to description on how to install ROS 2

wilcobonestroo:
The page describing how to install ROS 2 was moved.


wilcobonestroo:
On closer inspection, I am not sure where the link should actually point to. To me, it makes more sense to direct it to the latest version ([Humble](https://docs.ros.org/en/humble/Installation/Alternatives/Ubuntu-Install-Binary.html)) and not to rolling.

PR #334, opened 2022-07-18
Fix checker width from 200mm to 20mm

ijnek:
Hmm, i see that the checkerboard was 200mm squares in the [original tutorial](http://wiki.ros.org/camera_calibration/Tutorials/MonocularCalibration)

 I'm going to convert this PR to a draft for now..


SteveMacenski:
LGTM

PR #335, opened 2022-07-19
Fix missing --ros-args notation, and square parameter value

ijnek:


ijnek:
Would be good to squash the commits before merging

SteveMacenski (approved):

PR #343, opened 2022-09-06
Add note for delayed startups

ncdejito:
## Why
* For low compute, delays in startup makes processes start in the wrong order, resulting in malformed tf trees
* Resulting log messages direct towards solving different problems, but problems are only symptoms of this specific problem
* For beginners, prevent spending too much time guessing what the problem could be

## Symptom Errors
* Logs Timed out waiting for transform , frame does not exist
* RViz Global costmap No map received
* Message filter discarding message because queue is full

### DIFF
--- a/setup_guides/sensors/setup_sensors.rst
+++ b/setup_guides/sensors/setup_sensors.rst
@@ -505,11 +505,12 @@ Note that the parameters of the ``nav2_costmap_2d`` that we discussed in the pre
 
 After we have properly set up and launched Nav2, the ``/global_costmap`` and ``/local_costmap`` topics should now be active.
 
-Note: To make the costmaps show up, the 3 commands must be run in the right order
+.. note::  
+  To make the costmaps show up, run the 3 commands in this order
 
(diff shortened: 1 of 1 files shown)


setup_guides/sensors/setup_sensors.rst
```diff
@@ -505,6 +505,12 @@ Note that the parameters of the ``nav2_costmap_2d`` that we discussed in the pre
 
 After we have properly set up and launched Nav2, the ``/global_costmap`` and ``/local_costmap`` topics should now be active.
 
+Note: To make the costmaps show up, the 3 commands must be run in the right order
```
SteveMacenski, line 508: Generally LGTM but we should use the actual markdown `note` type so that it highlights. 

I'd also change the wording about "right", there is no "right" order, but just run in "this" order (due to DDS discovery and/or TF tree requirements of neighboring nodes)

ncdejito: done!

SteveMacenski (approved):
I really appreciate this note, there's been a few tickets I think that could have been avoided by this note. Thanks!

PR #344, opened 2022-09-07
update Costmap API documentation

JacksonK9:
update Costmap API documentation as the Pull Request of navigation2

* ros-planning/navigation2#3169

### DIFF
--- a/commander_api/index.rst
+++ b/commander_api/index.rst
@@ -147,6 +147,23 @@ This is a Python3 API for costmap 2d messages from the stack. It provides the ba
 +---------------------------------------+----------------------------------------------------------------------------+
 | getCostmapTimestamp()                 | Get costmap timestamp.                                                     |
 +---------------------------------------+----------------------------------------------------------------------------+
+| getCostXY(mx, my)                     | Get the cost (``int``) of a cell in the costmap using mx(``int``),         |
+|                                       | my(``int``) of Map Coordinate.                                             |
++---------------------------------------+----------------------------------------------------------------------------+
+| getCostIdx(index)                     | Get the cost (``int``) of a cell in the costmap using index(``int``)       |
++---------------------------------------+----------------------------------------------------------------------------+
+| setCost(mx, my, cost)                 | Set the cost (``np.int8``) of a cell in the costmap using mx(``int``),     |
+|                                       | my(``int``) of Map Coordinate.                                             |
++---------------------------------------+----------------------------------------------------------------------------+
+| mapToWorld(mx, my)                    | Get the wx(``float``), wy(``float``) of world coodinate XY using           |
+|                                       | mx(``int``), my(``int``) of map coordinate XY                              |
++---------------------------------------+----------------------------------------------------------------------------+
+| worldToMap(wx, wy)                    | Get the mx(``int``), my(``int``) of map coodinate XY using                 |
+|                                       | wx(``float``), wy(``float``) of world coordinate XY                        |
++---------------------------------------+----------------------------------------------------------------------------+
+| getIndex(mx, my)                      | Get the index(``int``) of the cell using mx(``int``), my(``int``) of       |
+|                                       | map coodinate XY                                                           |
++---------------------------------------+----------------------------------------------------------------------------+
 
 Examples and Demos
 ******************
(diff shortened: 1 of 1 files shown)


commander_api/index.rst
```diff
@@ -147,6 +147,23 @@ This is a Python3 API for costmap 2d messages from the stack. It provides the ba
 +---------------------------------------+----------------------------------------------------------------------------+
 | getCostmapTimestamp()                 | Get costmap timestamp.                                                     |
 +---------------------------------------+----------------------------------------------------------------------------+
+| getCostXY(mx, my)                     | Get the cost (``int``) of a cell in the costmap using mx(``int``),         |
+|                                       | my(``int``) of Map Coordinate.                                             |
++---------------------------------------+----------------------------------------------------------------------------+
+| getCostIdx(index)                     | Get the cost (``int``) of a cell in the costmap using index(``int``)       |
++---------------------------------------+----------------------------------------------------------------------------+
+| setCost(mx, my, cost)                 | Set the cost (``np.int8``) of a cell in the costmap using mx(``int``),     |
```
SteveMacenski, line 155: Why is this one np.int8 but others are just straight ints?

commander_api/index.rst
```diff
@@ -147,6 +147,23 @@ This is a Python3 API for costmap 2d messages from the stack. It provides the ba
 +---------------------------------------+----------------------------------------------------------------------------+
 | getCostmapTimestamp()                 | Get costmap timestamp.                                                     |
 +---------------------------------------+----------------------------------------------------------------------------+
+| getCostXY(mx, my)                     | Get the cost (``int``) of a cell in the costmap using mx(``int``),         |
```
SteveMacenski, line 150: Make the units similar to above, where they're in square brackets, ex

Get the cost [int] of a cell in the costmap using mx [int].

SteveMacenski: And make sure to have a space between the "thing" (mx, my, etc) and the units. Some are butted against each other

SteveMacenski (changes requested):

commander_api/index.rst
JacksonK9: The type of costmap is np.array of np.int8.
So I pointed that detail, But we can use just straight int.
Do you think I should change to Int? or Just leave it?

SteveMacenski: uint seems right, since negative values are invalid

JacksonK9: Okay I will change all the int type to np.uint8 and notify again.
Thank you for your comment 👍

JacksonK9:
I think document show the type with (``np.uint8``) and unit with [m].

So I add the type and unit to the document of commander api.

Thank you.

commander_api/index.rst
```diff
@@ -147,6 +147,23 @@ This is a Python3 API for costmap 2d messages from the stack. It provides the ba
 +---------------------------------------+----------------------------------------------------------------------------+
 | getCostmapTimestamp()                 | Get costmap timestamp.                                                     |
 +---------------------------------------+----------------------------------------------------------------------------+
+| getCostXY(mx, my)                     | Get the cost (``np.uint8``) of a cell in the costmap using mx (``int``)    |
```
SteveMacenski, line 150: again, units in square brackets.

SteveMacenski (changes requested):

commander_api/index.rst
```diff
@@ -147,6 +147,23 @@ This is a Python3 API for costmap 2d messages from the stack. It provides the ba
...
+|                                       | , my (``int``) of Map Coordinate.                                          |
++---------------------------------------+----------------------------------------------------------------------------+
+| getCostIdx(index)                     | Get the cost (``np.uint8``) of a cell in the costmap using index (``int``) |
```
SteveMacenski, line 153: Again, why are we mixing uints and ints?

SteveMacenski (changes requested):

commander_api/index.rst
SteveMacenski: Also, the actual API implementation uses `np.int8` not unsigned ints

JacksonK9: As I check these two lines.

setInitialPose(initial_pose)          | Sets the initial pose (``PoseStamped``) of the robot to localization.  

getResolution()                       | Get map resolution [m/cell].      

So I think the "variable type" is using parentheses and the "unit" is using square bracket.

As a result, I change like the below line.

mapToWorld(mx, my)                    | Get the wx (``float``) [m], wy (``float``) [m] of world coordinate XY using

Should I change to "Get the wx [``float``] [m], wy [[``float``] [m] of world coordinate XY using" ?

Can you tell me which one is better?

Thank you for your comment.

SteveMacenski (approved):

PR #353, opened 2022-10-14
add port

jwallace42:
Missing port for compute path to pose.

### DIFF
--- a/configuration/packages/bt-plugins/actions/ComputePathToPose.rst
+++ b/configuration/packages/bt-plugins/actions/ComputePathToPose.rst
@@ -89,7 +89,7 @@ Output Ports
   ============== =======
 
   Description
-    	Compute path to pose error code.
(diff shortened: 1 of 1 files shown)


configuration/packages/bt-plugins/actions/ComputePathToPose.rst
```diff
@@ -80,6 +80,17 @@ Output Ports
...
+:compute_path_to_pose_error_code:
+
+  ============== =======
+  Type           Default
+  -------------- -------
+  int16          N/A  
+  ============== =======
+
+  Description
+    	Compute path to pose error code.
```
SteveMacenski, line 92:
```suggestion
    	Compute path to pose error code. See ``ComputePathToPose`` action message for the enumerated set of error codes.
```

SteveMacenski (approved):

PR #358, opened 2022-10-18
faster debug

jwallace42:
Found faster debugging steps than the provided documentation.

### DIFF
--- a/tutorials/docs/get_backtrace.rst
+++ b/tutorials/docs/get_backtrace.rst
@@ -232,7 +232,20 @@ To debug directly from the nav2 bringup launch files you may want to do the foll
 .. note::
   Turning off composition has serious performance impacts. If this is important to you please follow "From Large Project". 
 
+Once your server crashes, you'll see a prompt like below in the xterm window. At this point you can now get a backtrace.
+
 .. code-block:: bash
+
+  (gdb)
+
+In this session, type ``backtrace`` and it will provide you with a backtrace.
+Copy this for your needs.
+See the example trace in the section above for an example.
+
+These traces take some time to get used to reading, but in general, start at the bottom and follow it up the stack until you see the line it crashed on.
+Then you can deduce why it crashed.
+When you are done with GDB, type ``quit`` and it will exit the session and kill any processes still up.
+It may ask you if you want to kill some threads at the end, say yes.
 
 Automatic backtrace on crash
 ============================
--- a/tutorials/docs/get_backtrace.rst
+++ b/tutorials/docs/get_backtrace.rst
@@ -223,7 +223,7 @@ From Nav2 Bringup
 
 To debug directly from the nav2 bringup launch files you may want to do the following:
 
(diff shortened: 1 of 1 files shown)


SteveMacenski:
Its not clear to me what about this is "faster"? Its also less generic, one intention I had for this was it was instructions you could use in any project in the ROS Ecosystem, just happens to be targeted to Nav2 for the purposes of demonstration. This would remove that and some important context of how to send parameters to these nodes in that situation.

Overall, I think its definitely more succinct, but it loses some generality I'd like to keep. 

Thoughts?

SteveMacenski:
https://github.com/ros-planning/navigation.ros.org/commit/42da2cbea610f82506ce5a0f7eab174b4c10cbce

I think this makes that intention a bit more explicitly clear

jwallace42:
> Its not clear to me what about this is "faster"? Its also less generic, one intention I had for this was it was instructions you could use in any project in the ROS Ecosystem, just happens to be targeted to Nav2 for the purposes of demonstration. This would remove that and some important context of how to send parameters to these nodes in that situation.
> 
> Overall, I think its definitely more succinct, but it loses some generality I'd like to keep.
> 
> Thoughts?

By "faster" I mean easier to set up. If we have to create a separate launch file that is more time consuming than just adjusting the current launch file. 

It is definitely less generic. It might be worth adding this to a separate section in the get_backtrack docs. I would like to have this in the docs to allow new contributors to debug quickly without having to worry about other launch files.

tutorials/docs/get_backtrace.rst
```diff
@@ -170,7 +170,7 @@ Then you can deduce why it crashed.
 When you are done with GDB, type ``quit`` and it will exit the session and kill any processes still up.
 It may ask you if you want to kill some threads at the end, say yes.
 
-From Nav2 Bringup
+From Large Project
 =================
```
SteveMacenski, line 174: The `=` length must match

tutorials/docs/get_backtrace.rst
```diff
@@ -218,6 +218,35 @@ Then you can deduce why it crashed.
 When you are done with GDB, type ``quit`` and it will exit the session and kill any processes still up.
 It may ask you if you want to kill some threads at the end, say yes.
 
+From Nav2 Bringup
+=================
+
+To debug directly from the nav2 bringup launch files you may want to do the following:
+
+- Add ``prefix=['xterm -e gdb -ex run --args']`` to the non-composed node.
```
SteveMacenski, line 226:
```suggestion
- Add ``prefix=['xterm -e gdb -ex run --args']`` to the non-composed node in the appropriate launch file. 
```

tutorials/docs/get_backtrace.rst
SteveMacenski, line 174:
```suggestion
==================
```

SteveMacenski (approved):

PR #359, opened 2022-10-18
updates

jwallace42:
Removed manual composition docs. The current documentation was referencing deleted files and the direction are out of date with [rolling](https://github.com/ros2/demos/blob/rolling/composition/src/manual_composition.cpp). 

I don't see a reason to add manual composition into the nav2 stack but if someone wants they could :).

### DIFF
--- a/setup_guides/lifecycle_composition/setup_lifecycle_composition.rst
+++ b/setup_guides/lifecycle_composition/setup_lifecycle_composition.rst
@@ -67,51 +67,11 @@ Composition
 Composition is another new concept in ROS 2 that was introduced to reduce the memory and CPU resources by putting multiple nodes in a single process. In Nav2, Composition can be used to compose all Nav2 nodes in a single process instead of launching them separately. This is useful for deployment on embedded systems where developers need to optimize resource usage.
 
 .. seealso::
-   More information on Composition can be found `here <https://docs.ros.org/en/galactic/Tutorials/Composition.html>`_.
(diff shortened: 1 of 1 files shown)


setup_guides/lifecycle_composition/setup_lifecycle_composition.rst
```diff
@@ -67,51 +67,11 @@ Composition
 Composition is another new concept in ROS 2 that was introduced to reduce the memory and CPU resources by putting multiple nodes in a single process. In Nav2, Composition can be used to compose all Nav2 nodes in a single process instead of launching them separately. This is useful for deployment on embedded systems where developers need to optimize resource usage.
 
 .. seealso::
-   More information on Composition can be found `here <https://docs.ros.org/en/galactic/Tutorials/Composition.html>`_.
+   More information on Composition can be found `here <https://docs.ros.org/en/rolling/Tutorials/Intermediate/Composition.html>`_.
 
-In the following subsections, we give an example on how to add a new Nav2 server, which we notionally call the ``route_server``, to our system. This can be done either through manual composition or dynamic composition.
+In the following section, we give an example on how to add a new Nav2 server, which we notionally call the ``route_server``, to our system. This can be done either through manual composition or dynamic composition.
```
SteveMacenski, line 72: Just delete all references to manual composition. That was a short-term thing we did before we could do dynamic composition with lifecycle nodes

SteveMacenski (changes requested):

SteveMacenski (approved):

PR #360, opened 2022-10-24
docs for changes to behavior server

jwallace42:
Documentation changes for adding a global costmap collision checker to the behavior server.

### DIFF
--- a/configuration/packages/configuring-behavior-server.rst
+++ b/configuration/packages/configuring-behavior-server.rst
@@ -16,7 +16,7 @@ Note: pre-Rolling/Humble this was the Recovery server, not behavior server. Laun
 Behavior Server Parameters
 **************************
 
-:costmap_topic:
+:local_costmap_topic:
 
   ============== ===========================
   Type           Default
@@ -25,9 +25,20 @@ Behavior Server Parameters
   ============== ===========================
 
   Description
-    Raw costmap topic for collision checking.
+    Raw costmap topic for collision checking on the local costmap.
 
-:footprint_topic:
+:global_costmap_topic:
+
+  ============== ===========================
+  Type           Default
+  -------------- ---------------------------
+  string         "global_costmap/costmap_raw"
+  ============== ===========================
+
+  Description
+    Raw costmap topic for collision checking on the global costmap.
+
+:local_footprint_topic:
 
   ============== ===================================
   Type           Default
@@ -36,7 +47,18 @@ Behavior Server Parameters
   ============== ===================================
 
   Description
-    Topic for footprint in the costmap frame.
+    Topic for footprint in the local costmap frame.
+
+:global_footprint_topic:
+
+  ============== ===================================
+  Type           Default
+  -------------- -----------------------------------
+  string         "global_costmap/published_footprint"
+  ============== ===================================
+
+  Description
+    Topic for footprint in the global costmap frame.
 
 :cycle_frequency:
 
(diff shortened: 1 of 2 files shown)


SteveMacenski (commented):
Assuming the original PR is right, this LGTM, but I think there will be some changes in the original PR or maybe even closed in general since an easier solution is just having 2 instances of the behavior server: one global and other local set with global/local costmap/footprint/frame settings.

migration/Humble.rst
```diff
@@ -66,27 +66,6 @@ The following errors codes are supported (with more to come as necessary): Unkno
 
 The following error codes are supported (with more to come as necessary): Unknown, TF Error, Invalid Path, Patience Exceeded, Failed To Make Progress, or No Valid Control.
 
-`PR #3251 <https://github.com/ros-planning/navigation2/pull/3251>`_ pipes the highest priority error code through the bt_navigator and defines the error code structure.
```
SteveMacenski, line 69: Deleted the piping docs?

migration/Humble.rst
```diff
@@ -107,6 +86,10 @@ SmootherSelector BT Node
 ************************
 `PR #3283 <https://github.com/ros-planning/navigation2/pull/3283>`_ adds a BT node to set the smoother based on a topic or a default. See the configuration guide :ref:`SimpleSmoother` for more details. 
 
-Publish Costmap Layers
```
SteveMacenski, line 110: Deleted layer publisher docs?

migration/Humble.rst
```diff
@@ -107,6 +86,10 @@ SmootherSelector BT Node
...
-**********************
-`PR #3320 <https://github.com/ros-planning/navigation2/pull/3320>`_ adds the ability for the nav2_costmap_2d package to publish out costmap data associated with each layer.
\ No newline at end of file
+Give Behavior Server Access to Both Costmaps
+********************************************
+`PR #3255 <https://github.com/ros-planning/navigation2/pull/3255>`_ addes the ability for a behavior to access the local and global costmap. 
+
+To update behaviors, any reference to the global_frame must be updated to the local_frame parameter
+along with the ``configuration`` method which now takes in the local and global collision checkers.
+Lastly, ``getResourceInfo`` must be overriden to return ``CostmapInfoType::LOCAL``.
```
SteveMacenski, line 95: Other options include GLOBAL (if uses theglobal costmap and/or footprint) or BOTH if both are required. This allows us to only create and maintain the minimum amount of expensive resources as required.

migration/Humble.rst
```diff
@@ -107,6 +107,16 @@ SmootherSelector BT Node
...
\ No newline at end of file
+`PR #3320 <https://github.com/ros-planning/navigation2/pull/3320>`_ adds the ability for the nav2_costmap_2d package to publish out costmap data associated with each layer.
+
+Give Behavior Server Access to Both Costmaps
+********************************************
+`PR #3255 <https://github.com/ros-planning/navigation2/pull/3255>`_ addes the ability for a behavior to access the local and global costmap. 
+
+To update behaviors, any reference to the global_frame must be updated to the local_frame parameter
+along with the ``configuration`` method which now takes in the local and global collision checkers.
+Lastly, ``getResourceInfo`` must be overriden to return ``CostmapInfoType::LOCAL``. Other options include ``GLOBAL``if the behavior useses global costmap and/or footprint)
```
SteveMacenski, line 121:
```suggestion
Lastly, ``getResourceInfo`` must be overriden to return ``CostmapInfoType::LOCAL``. Other options include ``GLOBAL`` if the behavior useses global costmap and/or footprint)
```

SteveMacenski (approved):

PR #361, opened 2022-10-26
Add "static_layer" to the local map plugins

FilippoGuarda:
the omission of static layer made impossible to generate and visualize the local map on rviz. Thanks to Giacomo Fanti @FantiG for finding the error

### DIFF
--- a/setup_guides/sensors/setup_sensors.rst
+++ b/setup_guides/sensors/setup_sensors.rst
@@ -373,7 +373,7 @@ In this subsection, we will show an example configuration of ``nav2_costmap_2d``
         height: 3
         resolution: 0.05
         robot_radius: 0.22
-        plugins: ["voxel_layer", "inflation_layer"]
+        plugins: ["voxel_layer", "inflation_layer", "static_layer"]
         voxel_layer:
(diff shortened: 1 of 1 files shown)


SteveMacenski:
This is not an accidental omission. While some applications will benefit from having a static layer in the local map, typically we discourage this since it introduces dependency on the trajectory planner with localization quality to transform the map. For example, if you robot becomes poorly localized, then things that are not actual obstacles will not be navigable by the trajectory planners because they're being shown as map obstacles rather than using only current sensor data readings.

That can be appropriate for robots with relatively good / consistent localization or if your robot lacks good coverage sensors, but this is not what I want to be the default behavior, intentionally.

setup_guides/sensors/setup_sensors.rst
```diff
@@ -373,7 +373,7 @@ In this subsection, we will show an example configuration of ``nav2_costmap_2d``
         height: 3
         resolution: 0.05
         robot_radius: 0.22
-        plugins: ["voxel_layer", "inflation_layer"]
+        plugins: ["voxel_layer", "inflation_layer", "static_layer"]
```
SteveMacenski, line 376: This is no entry for static_layer in the keys below.

PR #375, opened 2022-11-24
Update configuring-amcl.rst

borongyuan:
Fix recommended value of recovery_alpha_fast.

### DIFF
--- a/configuration/packages/configuring-amcl.rst
+++ b/configuration/packages/configuring-amcl.rst
@@ -285,7 +285,7 @@ Parameters
   ============== =============================
 
   Description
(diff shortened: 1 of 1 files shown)


SteveMacenski:
Can you explain this recommended value update?

borongyuan:
A good value of recovery_alpha_fast in the [code section](https://github.com/ros-planning/navigation2/blob/a392a17809178a1b1fd2bdbaca6204b579c5a81a/nav2_amcl/src/amcl_node.cpp#L166) is 0.1, as is the [ROS 1 documentation](http://wiki.ros.org/amcl). I guess the reason for this typo is that you forgot to modify it when copying from the recovery_alpha_slow part (#169).

SteveMacenski (approved):

PR #378, opened 2022-12-21
adding mppic docs

SteveMacenski:


configuration/packages/configuring-mppic.rst
```diff
@@ -0,0 +1,812 @@
...
+            offset_from_furthest: 5
+            threshold_to_consider: 0.6
+          PathAngleCritic:
+            enabled: true
+            cost_power: 1
+            cost_weight: 2.0
+            offset_from_furthest: 4
+            threshold_to_consider: 0.40
+            max_angle_to_furthest: 1.0
+          # TwirlingCritic:
```
padhupradheep, line 774: Maybe we need to explain it below, why this is commented? 

Also, I cannot find anywhere what this specific critic does?

SteveMacenski: Its commented out mostly because that's how its in my file :wink: I wanted the example to be complete (e.g. have all of the critics) but not necessarily the ones that are not enabled by default. It provides some context.

But this penalizes twirling motions - e.g. unnecessary rotations. It could potentially also help with maybe some of the angular velocity jitter, but its main goal is to add a small penalty so that omnidirectional robots don't start drifting in different rotational axes while moving. It discourages excessive rotations so you don't end up with a moving forward while also rotating a bit clockwise (because why not otherwise?)

plugins/index.rst
```diff
@@ -92,6 +92,10 @@ Controllers
 |                            |                    | variation on the pure pursuit    | Differential          |
 |                            |                    | algorithm with adaptive features.|                       |
 +----------------------------+--------------------+----------------------------------+-----------------------+
+| `MPPI Controller`_         | Steve Macenski     | A predictive MPC controller with | Differential, Omni,   |
```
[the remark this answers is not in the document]
SteveMacenski: A PR is always appreciated :wink: 

This is mostly stream of consciousness. Better to have something than nothing is my documentation motto

PR #382, opened 2023-01-06
Update index.rst

afifswaidan:
Added line iterator and collision checker documentation, also updated the docs for costmap_2d for the new function (``mapToWorldValidatged``).

### DIFF
--- a/commander_api/index.rst
+++ b/commander_api/index.rst
@@ -168,38 +168,8 @@ This is a Python3 API for costmap 2d messages from the stack. It provides the ba
 |                                       | map coordinate XY                                                          |
 +---------------------------------------+----------------------------------------------------------------------------+
 
-Line Iterator API
-*******************
-This is a Python3 API for a line iterator.
-It provides the ability to iterate
-through the points of a line, get coordinates, and check line length/validity.
-
-+---------------------------------------+----------------------------------------------------------------------------+
-| Line Iterator Method                  | Description                                                                |
-+=======================================+============================================================================+
-| isValid()                             | Check if line is valid.                                                    |
-+---------------------------------------+----------------------------------------------------------------------------+
-| advance()                             | Advance to the next point in the line.                                     |
-+---------------------------------------+----------------------------------------------------------------------------+
-| getX()                                | Get the abscissa of the current point.                                     |
-+---------------------------------------+----------------------------------------------------------------------------+
-| getY()                                | Get the ordinate of the current point.                                     |
-+---------------------------------------+----------------------------------------------------------------------------+
-| getX0()                               | Get the abscissa of the initial point.                                     |
-+---------------------------------------+----------------------------------------------------------------------------+
-| getY0()                               | Get the ordinate of the intial point.                                      |
-+---------------------------------------+----------------------------------------------------------------------------+
-| getX1()                               | Get the abscissa of the final point.                                       |
-+---------------------------------------+----------------------------------------------------------------------------+
-| getY1()                               | Get the ordinate of the final point.                                       |
-+---------------------------------------+----------------------------------------------------------------------------+
-| get_line_length()                     | Get the length of the line.                                                |
-+---------------------------------------+----------------------------------------------------------------------------+
-| clamp(n, min_n, max_n)                | Clamp n to be between min_n and max_n. (``float``)                         |
-+---------------------------------------+----------------------------------------------------------------------------+
-
 Footprint Collision Checker API
-*********************************
+*******************************
 This is a Python3 API for a Footprint Collision Checker.
 It provides the needed methods to manipulate the coordinates
 and calculate the cost of a Footprint in a given map.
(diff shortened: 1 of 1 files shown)


commander_api/index.rst
```diff
@@ -160,13 +160,71 @@ This is a Python3 API for costmap 2d messages from the stack. It provides the ba
...
-|                                       | wx (``float``) [m], wy (``float``) [m] of world coordinate XY              |
+| worldToMapValidated(wx, wy)           | Get the mx (``int``), my (``int``) of map coordinate XY using              |
+|                                       | wx (``float``) [m], wy (``float``) [m] of world coordinate XY.             |
+|                                       | If wx wy coordinates are invalid, (None,None) is returned.                 |
 +---------------------------------------+----------------------------------------------------------------------------+
 | getIndex(mx, my)                      | Get the index (``int``) of the cell using mx (``int``), my (``int``) of    |
 |                                       | map coordinate XY                                                          |
 +---------------------------------------+----------------------------------------------------------------------------+
 
+Line Iterator API
```
SteveMacenski, line 171: I think you can remove this section, since the line iterator is more an internal class for the collision detector

commander_api/index.rst
```diff
@@ -160,13 +160,71 @@ This is a Python3 API for costmap 2d messages from the stack. It provides the ba
...
++---------------------------------------+----------------------------------------------------------------------------+
+| getY1()                               | Get the ordinate of the final point.                                       |
++---------------------------------------+----------------------------------------------------------------------------+
+| get_line_length()                     | Get the length of the line.                                                |
++---------------------------------------+----------------------------------------------------------------------------+
+| clamp(n, min_n, max_n)                | Clamp n to be between min_n and max_n. (``float``)                         |
++---------------------------------------+----------------------------------------------------------------------------+
+
+Footprint Collision Checker API
+*********************************
```
SteveMacenski, line 202: The *** line has to match the length of the header (remove a couple)

commander_api/index.rst
```diff
@@ -160,13 +160,71 @@ This is a Python3 API for costmap 2d messages from the stack. It provides the ba
...
+This is a Python3 API for a Footprint Collision Checker.
+It provides the needed methods to manipulate the coordinates
+and calculate the cost of a Footprint in a given map.
+
++----------------------------------------------+--------------------------------------------------------------------------------------------+
+| Footprint Collision Checker Method           | Description                                                                                |
++==============================================+============================================================================================+
+| footprintCost(footprint)                     | Iterate over all the points in a footprint (``Polygon``) and check for collision.          |
++----------------------------------------------+--------------------------------------------------------------------------------------------+
+| lineCost(x0, x1, y0, y1, step_size=0.1)      | Iterate over all the points along a line and check for collision.                          |
```
SteveMacenski, line 212: Is the default 0.1? Shouldn't it be 1 cell?

commander_api/index.rst
```diff
@@ -160,13 +160,71 @@ This is a Python3 API for costmap 2d messages from the stack. It provides the ba
...
+Footprint Collision Checker API
+*********************************
+This is a Python3 API for a Footprint Collision Checker.
+It provides the needed methods to manipulate the coordinates
+and calculate the cost of a Footprint in a given map.
+
++----------------------------------------------+--------------------------------------------------------------------------------------------+
+| Footprint Collision Checker Method           | Description                                                                                |
++==============================================+============================================================================================+
+| footprintCost(footprint)                     | Iterate over all the points in a footprint (``Polygon``) and check for collision.          |
```
SteveMacenski, line 210: Id just say "Checks the footprint (``Polygon``) for collision at its implicit provided coordinate pose."

The iterate over points thing lacks the detail about the line iteration stuff, which I think is TMI for this 1-liner

SteveMacenski (changes requested):

commander_api/index.rst
afifswaidan: We made it default to 1 cell in the Line Iterator API, but here this function is being called inside the ```footprintCost``` function. Since footprints usually aren't very big the lines in it (in the polygon) can be iterated quickly with a step size of 0.1m.
Still, the user can increase the step size if the footprint is big enough.

commander_api/index.rst
afifswaidan: Okay, will do that.

commander_api/index.rst
afifswaidan: I just noticed that also in the previous APIs, the *** line always 2 (*) characters more than the header.
Thats why I made it longer by 2 (*). I will remove them.
Thanks for the notice!

commander_api/index.rst
afifswaidan: Okay sure, will do that.

commander_api/index.rst
afifswaidan: Okay, I got your point.
If we aim for speed more than just accuracy then definitely we should increase the step size.
Should I make it default to 1.0 or 0.5 in the ``lineCost`` function?

SteveMacenski: 0.5 seems like a reasonable default

SteveMacenski (approved):

PR #398, opened 2023-03-07
Fix Typos in docs

tonynajjar:


SteveMacenski (approved):

SteveMacenski:
I’ll fully admit much of this documentation is never re-read before publication. I always figure some (buggy) documentation is far better than nothing or waiting until I think its perfect before putting into the hands of users 🙃

PR #400, opened 2023-03-11
add Tailos' rosie robot https://tailos.com

AustinDeric:


about/robots.rst
```diff
@@ -19,6 +19,8 @@ Click on the images below for a link to the drivers or navigation configurations
 +------------------------+------------------------+------------------------+
 |         |tb3|          +         |rover|        +           |tb2|        |
 +------------------------+------------------------+------------------------+
+|         |rosie|          +         ||        +           ||        |
```
SteveMacenski, line 22: Looking at the builds, this isn't working (see artifact of build: https://output.circle-artifacts.com/output/job/2fbc4e63-aa28-48ec-9f11-f30149bf70c5/artifacts/0/html/about/robots.html). The table isn't there. 

Try filling the other 2 fields with the `|soon|` field. Also, the table needs to be properly formatted. Markdown doesn't care about it, but restructured text does :smile: See the other rows above this one for examples!

SteveMacenski (changes requested):
Thanks so much for adding this! Great to see another user of this work!

about/robots.rst
AustinDeric: done, ty!

AustinDeric:
> Thanks so much for adding this! Great to see another user of this work!

yea we are looking at increasing our contributions back to the community this year as well.  here are some recent ones that come to mind: 
1. https://github.com/tailosinc/ros2_nav2/pull/24
2. https://github.com/SteveMacenski/slam_toolbox/pull/505

about/robots.rst
```diff
@@ -19,6 +19,8 @@ Click on the images below for a link to the drivers or navigation configurations
 +------------------------+------------------------+------------------------+
 |         |tb3|          +         |rover|        +           |tb2|        |
 +------------------------+------------------------+------------------------+
+|         |rosie|        +         |soon|        +           |soon|        |
```
SteveMacenski, line 22: Still 1 space off here, it won't render :frowning:

SteveMacenski:
Your (1) is a private repo, so I can't see it :wink: 

> yea we are looking at increasing our contributions back to the community this year as well. If you have any ideas for this, i'd be more than happy to discuss further. Would be great if you could meet our world-class team.

That's really great to hear! Do you have an email address I can contact you over? I have a couple of things that might be worth discussing. I'd also be happy to meet the team!

PR #407, opened 2023-03-27
Add docs for Dev Containers

ruffsl:
Related:
- https://github.com/ros-planning/navigation2/pull/3482

TODO:

- [x] markdown -> restructuredtext
- [x] refactor docker build instructions


ruffsl:
@SteveMacenski , here is some documentation on my devcontainer work that we discussed over the previous Nav2 meeting.

ruffsl:
We can save the restructuredtext for last, and I'll refactor docker build instructions based on feedback of what is written thus far.

build_instructions/devcontainers.md
```diff
@@ -0,0 +1,160 @@
+# Dev Containers
```
SteveMacenski, line 1: I think this needs to be linked from the index page to be findable. Maybe also in `rst` format? But to be honest, I don't know a ton about sphinx, so it wouldn't surprise me if md files were fine too

SteveMacenski (commented):
So this isn't strictly for codespaces? I read through most of it and seems good to me! 

The only thing I'd ask is that if you haven't yet, to start with a blank slate and follow only the letter of your instructions in this page to see if you can get a successful result at the end. If not, add the details that might be missing for a novice user. We don't want this page to become a hot spot of tickets asking for support because something was unclear or they got stuck on something obvious to you as an existing user of the capability

ruffsl:
Starting as 1c3996f , I've refactored the build and docker instructions.

As this PR is getting rather large, I think I'll split the devcontainer visualization docs to a seperate PR.

.circleci/config.yml
```diff
@@ -17,7 +17,7 @@ references:
         apt install openjdk-8-jre -y
         apt install bash git openssh-server -y
         apt install python3-pip python3 -y
-        pip3 install sphinx==3.5.0 docutils==0.14 sphinx_rtd_theme breathe==4.28.0 sphinxcontrib-plantuml jinja2==3.0.3
+        pip3 install sphinx==3.5.0 docutils==0.14 sphinx_rtd_theme breathe==4.28.0 sphinxcontrib-plantuml jinja2==3.0.3 myst-parser sphinx_copybutton
```
SteveMacenski, line 20: Set a version number? Also in the readme.md / docker file

index.rst
```diff
@@ -137,6 +137,7 @@ Below is an example of the TB3 navigating in a small lounge.
 
    getting_started/index.rst
    build_instructions/index.rst
+   dev_containers/index.md
```
SteveMacenski, line 140: Perhaps this should be under build instructions as an option?

build_instructions/index.rst
```diff
@@ -9,126 +9,133 @@ Install
...
-.. code:: bash
+* Rolling Development Source
 
-  mkdir -p ~/nav2_ws/src
-  cd ~/nav2_ws/src
-  git clone https://github.com/ros-planning/navigation2.git --branch <ros2-distro>-devel
-  cd ~/nav2_ws
-  rosdep install -y -r -q --from-paths src --ignore-src --rosdistro <ros2-distro>
-  colcon build --symlink-install
+  * Intricately build Nav2 using from latest source dependencies
```
SteveMacenski, line 31: Why is this "intricately"? Just say "Build..."

build_instructions/index.rst
```diff
@@ -9,126 +9,133 @@ Install
...
 
-  mkdir -p ~/nav2_ws/src
-  cd ~/nav2_ws/src
-  git clone https://github.com/ros-planning/navigation2.git --branch main
-  cd ~/nav2_ws
-  rosdep install -y -r -q --from-paths src --ignore-src --rosdistro rolling
-  colcon build --symlink-install
-  source install/setup.bash
+.. attention::
+   For ROS 2 distros prior to Galactic, the branch naming schema was ``<distro>-devel``, while the default branch for Rolling is ``main``.
```
SteveMacenski, line 56: Remove, those are all EOL so we don't need to provide support anymore

build_instructions/index.rst
```diff
@@ -9,126 +9,133 @@ Install
...
 
-You are now ready for the demonstrations!
+.. code:: bash
 
-Docker
-******
+  source /opt/ros/<distro>/setup.bash
+  export NAV2_WS=~/nav2_ws
```
SteveMacenski, line 61: remove this export and just make `nav2_ws` in the directory, lets not complicate things for novice users :smile:

build_instructions/index.rst
```diff
@@ -9,126 +9,133 @@ Install
...
+  mkdir -p $NAV2_WS/src && cd $NAV2_WS
```
SteveMacenski, line 62:
```suggestion
  mkdir -p $NAV2_WS/src && cd $NAV2_WS/src
```

build_instructions/index.rst
```diff
@@ -9,126 +9,133 @@ Install
...
+  git clone https://github.com/ros-planning/navigation2.git --branch $ROS_DISTRO ./src/navigation2
```
SteveMacenski, line 63:
```suggestion
  git clone https://github.com/ros-planning/navigation2.git --branch $ROS_DISTRO 
```

build_instructions/index.rst
```diff
@@ -9,126 +9,133 @@ Install
...
+  rosdep install -y \
+    --from-paths ./src \
```
SteveMacenski, line 65:
```suggestion
    --from-paths . \
```

build_instructions/index.rst
```diff
@@ -9,126 +9,133 @@ Install
...
+    --ignore-src
+  colcon build \
```
SteveMacenski, line 67: add in a `cd ../../`

build_instructions/index.rst
```diff
@@ -9,126 +9,133 @@ Install
...
 
-It is though generally recomended to install Nav2 releases from the apt repository inside a container if you'd like to use our released binaries.
+.. hint::
+  For more examples on building Nav2 from released distribution binaries, checkout `distro.Dockerfile <https://github.com/ros-planning/navigation2/blob/main/tools/distro.Dockerfile>`_.
 
 .. rst-class:: content-collapse
 
-Building Docker Container
-=========================
+Rolling Development Source
```
SteveMacenski, line 77: Repeat above here too :smile:

build_instructions/index.rst
```diff
@@ -9,126 +9,133 @@ Install
...
+  docker build --tag navigation2:$ROS_DISTRO \
+    --build-arg FROM_IMAGE=ros:$ROS_DISTRO \
+    --build-arg OVERLAY_MIXINS="release ccache lld" \
+    --cache-from ghcr.io/ros-planning/navigation2:main \
+    ./navigation2
+
+The `docker build <https://docs.docker.com/engine/reference/commandline/build/>`_ command above creates a tagged image using the `Dockerfile` from the context specified using the path to the repo, where build-time variables are set using additional arguments, e.g. passing a set of `colcon mixins <https://github.com/colcon/colcon-mixin-repository>`_ to configure the workspace build. Check the ``ARG`` directives in the `Dockerfile` to discover all build-time variables available. The command also specifies an `external cache source <https://docs.docker.com/engine/reference/commandline/build/#cache-from>`_ to pull the latest cached image from Nav2's `Container Registry <https://github.com/ros-planning/navigation2/pkgs/container/navigation2>`_ to speed up the build process.
+
+.. tip::
+  The images cached from above are used for Nav2 CI, but can also be used with Nav2 :ref:`dev-containers`!
```
SteveMacenski, line 138: It seems natural to add a new drop down in this section for the Dev Container / Codespaces instead of having it as a new root-website section. That dropdown box like the others can link to the appropriate subpages

dev_containers/dev_containers_docs/development_guide.md
```diff
@@ -0,0 +1,191 @@
+# Development Guide
```
SteveMacenski, line 1: I think the link to this page from the index page should be more prominent. It looks like another "just for context" link but its actually quite important to make sure folks read.

.circleci/config.yml
ruffsl: 7a536f8df6db23abd2d7d563071f322ae6a1ba1f ... f0e2f3f8fb98dbaae3ec91d853fe94a421f24eab

build_instructions/index.rst
ruffsl: It serves as an adverb to briefly describe its purpose and contrast the involvement with the other methods, E.g. one build method is simple, one is extensive, and one is fast. There's no free lunch, so the reader can quickly assess what bed to make and lie in. I'm open to better adjective descriptors. Perhaps `exhaustively` or `rigorously`?

build_instructions/index.rst
ruffsl: I'd actually recommend against changing the working directory more than needed. When thinking of novice users, it's more instructive to be explicit with the paths, i.e. which paths are used by which commands, rather than than abbreviating to relative dot paths, that are easier to misread, miscopy, or accidently reorder, either from font size or fat fingers. This is the same reason why full name arguments are used, rather than their terse abbreviations.

For example, for a long time, I though `src` in `vcs import src < ...` was a subcommand, rather than a path, as the order of operations and command line parsing was not obvious or self descriptive.

Another common example, is that students I teach will often invoke `colcon build` from inside the `src` or `git clone`ed folder, after forgetting to cd back out to the root of the colcon workspace. Colcon will (frustratingly) play along just fine, and add the build, install and log folders directly inside their repos.

build_instructions/index.rst
ruffsl: https://github.com/ros-planning/navigation.ros.org/pull/407#discussion_r1170053805

build_instructions/index.rst
ruffsl: https://github.com/ros-planning/navigation.ros.org/pull/407#discussion_r1170053805

build_instructions/index.rst
ruffsl: https://github.com/ros-planning/navigation.ros.org/pull/407#discussion_r1170053805

build_instructions/index.rst
ruffsl: The devcontainer root is intended to be a more general introduction, index, and resource reference. For example, that root is where I'm attaching the added docs on visualizing Nav2 when using dev containers. Concentrating all of that info into another build drop down would be quite out of scope for this sub index.

- https://github.com/ruffsl/navigation.ros.org/pull/1

Aside from the devcontainer CLI, there isn't a way to build the devcontainer from the command line using VS Code:

- https://github.com/microsoft/vscode-remote-release/issues/2133

While most of the best ergonomics in using dev containers is provided when using the Remote VS Code extensions, there isn't really a command line play-by-play to include, and would be better illustrated by just referencing the introduction.

dev_containers/dev_containers_docs/development_guide.md
ruffsl: Do you know of another way to reference this guide in sphinx? It looks rather fine/clean from the sidebar or toctree:
image

And is linked in the body of the root index as well:

https://github.com/ruffsl/navigation.ros.org/blob/506f01254fb4553ad418792f33f2abc1a52ea763/dev_containers/index.md#L83

build_instructions/index.rst
SteveMacenski: > Colcon will (frustratingly) play along just fine

That did cause me lots of problems right when ROS 2 betas came out. It would work fine in ROS 1 but _not_ in ROS 2 very subtly. 

I can respect that opinion. When trying to figure out why I feel like that is a better way for novices, the only thing I can point to is that ROS 2 documentation for builds recommend that pattern https://docs.ros.org/en/foxy/Installation/Alternatives/Ubuntu-Development-Setup.html so I suppose I've just implicitly mimicked it. I don't feel overly strong about it, but there is some value in consistency to a user experience across the ecosystem.

build_instructions/index.rst
SteveMacenski: I just don't think this belongs in the root navigating bar on the website and this is related to building and development which already has a root-level page. I think in some way it should be combined with that more fully. 

I don't think you should put all that info in the drop down, certainly! I think just having that be the root entry point for your subpages that is and remove the root-navigation-bar entries for it. Or don't use the drop down and have a new section for dev containers within the build & install page. I'm not really picky and leave it to your discretion, but I want to keep that root navigation bar as small as possible, its already quite long

build_instructions/index.rst
ruffsl: > Why is building nav2 any harder for Rolling than it is for Foxy?

It's not really. The only distinction I tried to preserve with the refactoring was whether the user wanted to know how to build from released binaries, or build from the source dependencies. Those source dependencies could be Rolling or Foxy, I just titled it as Rolling, as most are only ever going to build the ROS 2 Rolling branches from source.

> everything below Nav2 is available in binary format in all distributions so colcon build for the Nav2 repo is equally easy for both

Yes, but in the past we've had to build our underlay from source as well. Particularly during turbulent transitions between ROS and Ubuntu distributions. I think it still handy to document how to build development dependencies from scratch if need be.

SteveMacenski: True, but I think those days are over. We're now at Indigo-number of releases, which in ROS 1 didn't really have any major issues like that in the future. I'd also point out that our repos file is right now empty (commented out in case needed) so it will be getting everything from binary format.

I don't want subjective qualifiers on the ease of steps to influence what distributions users use. If you make it seem like its significantly harder (even though its just 1 command), that will detract users

ruffsl: > True, but I think those days are over.

Famous last words... ~~Winter~~ Ubuntu 24.04 is coming... ❄️☃️ 

> I'd also point out that our repos file is right now empty (commented out in case needed) so it will be getting everything from binary format.

Seriously though, it can still be handy to build using custom dependencies, like repo forks of eigan or OpenCV, etc. When I was working on SLAM frameworks, not a day went by that I'd have to resort to some odd branch of OpenCV or PCL.

> Simply put, I don't want subjective qualifiers on the ease of steps to influence what distributions users use.

That's fair. I've revised it to just be descriptive instead: 0cb6fb3a2cd3f849335ffffffa896a17d2836da4

SteveMacenski: Definitely, that's why the repos file still exists. As we pull in new algorithms, its useful for us to be able to move fast in CI even while rosdistro is catching up!

build_instructions/index.rst
ruffsl: Looks like the ROS 2 documentation for builds do the same thing we do here, and elect to never cd into the src or cloned repos, but only to the root of the colcon workspace. We merely git clone from the same directory they would vcs import.

dev_containers/dev_containers_docs/development_guide.md
ruffsl: I played around with using implicit targets, but they didn't inherently produce anything more visually highlighted or prominent as compared to boldened hyperlinks using markdown:

https://myst-parser.readthedocs.io/en/latest/syntax/cross-referencing.html#implicit-targets

I then tried increasing the hyperlink font size by using sub headers, but they had the undesirable effect of duplicating subsect titles nested in the side bare. I also think re-using any heading typfaces an implicit hyperlink would be unwise from a consistent typography point of view.

I think the current approach matches the embellished pattern listing subsections using bullet point in the body of the section's index, as used elsewhere:

-  tutorials/index.rst
- configuration/index.rst
- plugin_tutorials/index.rst

Although, myst doesn't seem to have the one-to-one directive for `..toctree::`, so I've just bullet pointed them with summaries. But as I add more subsections using separate pages, is seem natural to index them using list just as the reader concludes with the getting started section.

build_instructions/index.rst
ruffsl: https://github.com/ros-planning/navigation.ros.org/pull/407#discussion_r1170308330

build_instructions/index.rst
ruffsl: > Or don't use the drop down and have a new section for dev containers within the build & install page.

I thought of doing that, but additional dev container subpages for Nav2, such as the Visualisation Guide or possible Codespace Guide, wouldn't fit the narrow category of "build and install". We could rename "build_instructions" subpage to "development_instructions" or something? Then keep the Build and Install heading in that section's root index.

ruffsl: See commits starting from e6aa6e4300f89d997020088d73f560692a5148c5 where I moved around the build instruction and dev container indexs into common root subject. I also moved the contribution into the same root in 265c2cb037f8ca18e273bcdf20d0af66431bd6f8 , as it seemed relevant, and could help shorten the root navigation bar further.

SteveMacenski:
It looks like your changes broke the mechanism where artifacts are generated so I can see the output and website files in circleci https://app.circleci.com/pipelines/github/ros-planning/navigation.ros.org/1676/workflows/5e58a602-fac9-421b-993b-abd9d0e0dfaa/jobs/1985/artifacts. The job has no recorded artifacts.

.circleci/config.yml
```diff
@@ -3,25 +3,26 @@ version: 2.1
 references:
   on_checkout: &on_checkout
     checkout:
-      path: nav2_docs
+      path: .
```
SteveMacenski, line 6: I assume these directory changes have something to do with it

ruffsl: Fixed with c95f117 .

That's really nifty how the relative hyperlinks from HTML artifacts still work when hosted from CircleCI.

PR #413, opened 2023-04-08
Fix rviz exec dependency in documentation.

ValerioMa:
Easy to confuse between ros 1 and ros 2  but the correct package for ros 2 is rviz2


SteveMacenski:
No it is not https://github.com/ros/rosdistro/blob/master/rolling/distribution.yaml#L5477 We need the full metapackage

PR #416, opened 2023-04-18
add m220-av to robots using

Aposhian:


SteveMacenski (approved):

SteveMacenski:
https://github.com/ros-planning/navigation.ros.org/pull/425

This includes your robot, thanks! Updated the table and images so things are easier to add in the future

PR #417, opened 2023-04-29
Add Visualization docs for Dev Containers

ruffsl:
Related:
- https://github.com/ros-planning/navigation2/pull/3523
- https://github.com/ros-planning/navigation.ros.org/pull/407

TODO: Document
- [ ] Web Apps
- [ ] Native Apps
- [ ] Desktop Stream
- [ ] Display Socket


SteveMacenski:
I'm not sure you totally need to provide so much context on the options, but I suppose showing / explaining using rviz + gazebo would be nice. I think going into AWS / Construct / other options that don't have direct parallels for users is maybe a bit too much context?

PR #418, opened 2023-05-03
Remove tb3 manual install

Ryanf55:
* The necessary deps are now installed when you install nav2_bringup, or use rosdep
* Follow up to https://github.com/ros-planning/navigation2/pull/3556

### DIFF
--- a/getting_started/index.rst
+++ b/getting_started/index.rst
@@ -32,7 +32,7 @@ Installation
 
    .. code-block:: bash
 
-      sudo apt install ros-<ros2-distro>-turtlebot3*
+      sudo apt install ros-<ros2-distro>-turtlebot3-gazebo
 
 Running the Example
 *******************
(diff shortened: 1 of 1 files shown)


getting_started/index.rst
```diff
@@ -28,12 +28,6 @@ Installation
       sudo apt install ros-<ros2-distro>-navigation2
       sudo apt install ros-<ros2-distro>-nav2-bringup
 
-3. Install the Turtlebot 3 packages:
```
SteveMacenski, line 31: Readd and change to "3. Install the Turtlebot 3 packages (Humble and older):" 

This still needed for non-main distributions

Ryanf55: Ah, right on. Fixed

While I'm updating the docs, should I also change the install to just install 
`ros-<ros2-distro>-turtlebot3-gazebo`?
This will save users space. 

I searched the code for turtlebot and couldn't find anything obvious that seemed to need anything except these resources. Easy enough to test in CI.

SteveMacenski: I think that is correct

Ryanf55: Changed in 9b7d62c9b5fc1038da0d5821bb17daf785f4771b

SteveMacenski (approved):

SteveMacenski:
Thanks!

PR #424, opened 2023-05-22
viz_expansions planner parameter documentation

doisyg:
For https://github.com/ros-planning/navigation2/pull/3577

### DIFF
--- a/configuration/packages/smac/configuring-smac-hybrid.rst
+++ b/configuration/packages/smac/configuring-smac-hybrid.rst
@@ -220,7 +220,7 @@ Parameters
   ====== =======
 
   Description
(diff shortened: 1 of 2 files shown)


configuration/packages/smac/configuring-smac-hybrid.rst
```diff
@@ -211,6 +211,17 @@ Parameters
...
+:``<name>``.viz_expansions:
+
+  ====== =======
+  Type   Default                                                   
+  ------ -------
+  bool   false         
+  ====== =======
+
+  Description
+    Whether to publish expansions on the /expansions topic as an array of poses (the orientation has no meaning). WARNING: heavy to compute and to display, for debug only as it degrades the performance.
```
SteveMacenski, line 223:
```suggestion
    Whether to publish expansions on the ``/expansions`` topic as an array of poses (the orientation has no meaning). WARNING: heavy to compute and to display, for debug only as it degrades the performance.
```

PR #438, opened 2023-06-20
Option for ObstacleLayer to not overwrite StaticLayer's unknown parts

RBT22:
The documentation for [#3612](https://github.com/ros-planning/navigation2/pull/3612)

### DIFF
--- a/configuration/packages/costmap-plugins/obstacle.rst
+++ b/configuration/packages/costmap-plugins/obstacle.rst
@@ -49,7 +49,18 @@ Obstacle Layer Parameters
   ====== =======
 
   Description
-    Enum for method to add data to master costmap, default to maximum.
+    Enum for method to add data to master costmap. Must be 0, 1 or 2, default to 1 (see below).
+
+  | 0 - Overwrite: Overwrite master costmap with every valid observation.
+
+  | 1 - Max: Sets the new value to the maximum of the master_grid's value and this layer's value. 
+  | If the master value is NO_INFORMATION, it is overwritten. This is the default.
+  
+  | 2 - MaxWithoutUnknownOverwrite: Sets the new value to the maximum of the master_grid's
+  | value and this layer's value. If the master value is NO_INFORMATION, it is NOT overwritten.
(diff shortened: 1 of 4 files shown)


RBT22:
> Need to add to migration guide on website that this new feature exists to folks can use it!

Where should I put the migration guide?

> Need to add on configuration guide for costmap 2d the new setting so folks in the future can know its there

What should be this one? Isn't it enough to put info about it in the layer params docs?

configuration/packages/costmap-plugins/obstacle.rst
```diff
@@ -49,7 +49,18 @@ Obstacle Layer Parameters
   ====== =======
 
   Description
-    Enum for method to add data to master costmap, default to maximum.
+    Enum for method to add data to master costmap. Must be 0, 1 or 2, default to 1 (see below).
+
+  | 0 - Overwrite: Overwrite master costmap with every valid observation.
+
+  | 1 - Max: Sets the new value to the maximum of the master_grid's value and this layer's value. 
+  | If the master value is NO_INFORMATION, it is overwritten. This is the default.
```
SteveMacenski, line 57:
```suggestion
  | This is the default.
```

configuration/packages/costmap-plugins/voxel.rst
```diff
@@ -104,7 +104,17 @@ Voxel Layer Parameters
   ====== =======
 
   Description
-    Enum for method to add data to master costmap, default to maximum.
+    Enum for method to add data to master costmap. Must be 0, 1 or 2, default to 1 (see below).
+
+  | 0 - Overwrite: Overwrite master costmap with every valid observation.
+
+  | 1 - Max: Sets the new value to the maximum of the master_grid's value and this layer's value. 
+  | If the master value is NO_INFORMATION, it is overwritten. This is the default.
```
SteveMacenski, line 112:
```suggestion
  | This is the default.
```

SteveMacenski (changes requested):
We also need an entry in the migration guide announcing this change

SteveMacenski (approved):

PR #453, opened 2023-08-04
Fix Galactic.rst (link to wrong pull request)

ymd-stella:
It seems to be unintentionally the same as one link above.


SteveMacenski (approved):

SteveMacenski:
Huh, odd, thanks!

PR #460, opened 2023-09-05
[binary filter] change boolean ros params documentation

enricosutera:
This PR integrates docs for https://github.com/ros-planning/navigation2/pull/3796, expanding binary filter configuration guide.

### DIFF
--- a/configuration/packages/costmap-plugins/binary_filter.rst
+++ b/configuration/packages/costmap-plugins/binary_filter.rst
@@ -3,7 +3,7 @@
 Binary Filter Parameters
 ========================
 
-Binary Filter - is a Costmap Filter that publishes a boolean topic, flipping binary state when the encoded filter space value (corresponding to the filter mask point where the robot is) is higher than given threshold. It then flips back when lower or equal.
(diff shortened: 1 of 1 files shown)


configuration/packages/costmap-plugins/binary_filter.rst
```diff
@@ -3,7 +3,7 @@
 Binary Filter Parameters
 ========================
 
-Binary Filter - is a Costmap Filter that publishes a boolean topic, flipping binary state when the encoded filter space value (corresponding to the filter mask point where the robot is) is higher than given threshold. It then flips back when lower or equal.
+Binary Filter - is a Costmap Filter that publishes a boolean topic, flipping binary state when the encoded filter space value (corresponding to the filter mask point where the robot is) is higher than given threshold. It then flips back when lower or equal. Along with publishing the state, the filter can change boolean ros parameters of other ros nodes. The parameters to be changed can be configured as ros paramters themselves.
```
SteveMacenski, line 6:
```suggestion
Binary Filter - is a Costmap Filter that publishes a boolean topic, flipping binary state when the encoded filter space value (corresponding to the filter mask point where the robot is) is higher than given threshold. It then flips back when lower or equal. Along with publishing the state, the filter can change boolean ros parameters of other ros nodes. The parameters to be changed can be configured as ros parameters themselves. 
```

SteveMacenski (changes requested):
Needs migration guide entry too on the change

enricosutera:
> Needs migration guide entry too on the change

https://github.com/enricosutera/navigation.ros.org/commit/78a9bcb95832bde4fd75f0dc1f91993ea01cfb4c

PR #477, opened 2023-10-12
Update docu for PR #3866

kfabian:
Add new parameter and migration guide for PR #3866

### DIFF
--- a/configuration/packages/configuring-costmaps.rst
+++ b/configuration/packages/configuring-costmaps.rst
@@ -220,7 +220,7 @@ Costmap2D ROS Parameters
   ============== =======
 
   Description
-    Time to wait for the transform from robot base frame to global frame to become available.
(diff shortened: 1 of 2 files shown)


configuration/packages/configuring-costmaps.rst
```diff
@@ -211,6 +211,17 @@ Costmap2D ROS Parameters
...
+:initial_transform_timeout:
+
+  ============== =======
+  Type           Default
+  -------------- -------
+  double         60.0   
+  ============== =======
+
+  Description
+    Time to wait for the transform from robot base frame to global frame to become available.
```
SteveMacenski, line 223:
```suggestion
    Time to wait for the transform from robot base frame to global frame to become available. If exceeded, the  configuration stage is aborted.
```

SteveMacenski (approved):

PR #490, opened 2023-11-17
Add documentation for wait_for_service_timeout #3960

embeddedadam:


configuration/packages/configuring-bt-navigator.rst
```diff
@@ -111,6 +111,18 @@ Parameters
...
+:wait_for_service_timeout:
+
+  ==== =======
+  Type Default
+  ---- -------
+  int  1000
+  ==== =======
+
+  Description
+    Default timeout value (in milliseconds) while a BT nodes are waiting for acknowledgement from an service server.
```
SteveMacenski, line 123:
```suggestion
    Default timeout value (in milliseconds) while Action or Service BT nodes will waiting for acknowledgement from an service or action server on BT initialization (e.g. ``wait_for_action_server(timeout)``).
```

PR #503, opened 2023-12-23
Document optional TwistStamped support

Ryanf55:
To be merged with https://github.com/ros-planning/navigation2/pull/3775.

Many of the behaviors inherit from TimedBehavior. Is it worth documenting the same parameter in all of them, or is there a better place to reduce docs duplication?

### DIFF
--- a/configuration/packages/configuring-behavior-server.rst
+++ b/configuration/packages/configuring-behavior-server.rst
@@ -226,6 +226,19 @@ Spin distance is given from the action request
   Description
     maximum rotational acceleration (rad/s^2).
 
+:enable_stamped_cmd_vel:
+
+  ============== =============================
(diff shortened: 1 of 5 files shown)


configuration/packages/configuring-behavior-server.rst
```diff
@@ -258,6 +271,19 @@ DriveOnHeading distance, speed and time_allowance is given from the action reque
   Description
     Time to look ahead for collisions (s).
 
+:enable_stamped_cmd_vel:
```
SteveMacenski, line 274: Missing backup

SteveMacenski (changes requested):
Needs migration guide entry since this is very important. Place this near the top of the page

PR #517, opened 2024-02-05
Documentation for Graceful Motion

ajtudela:
Documentation for [#4021](https://github.com/ros-planning/navigation2/pull/4021)

### DIFF
--- a/configuration/packages/configuring-graceful-motion-controller.rst
+++ b/configuration/packages/configuring-graceful-motion-controller.rst
@@ -1,0 +1,232 @@
+.. _configuring_graceful_motion_controller:
+
+Graceful Motion Controller
+##########################
+
+Source code on Github_.
+
+.. _Github: https://github.com/ros-planning/navigation2/tree/main/nav2_graceful_motion_controller
+
+The graceful motion controller implements a controller based on the works of Jong Jin Park and Benjamin Kuipers in "A Smooth Control Law for Graceful Motion of Differential Wheeled Mobile Robots in 2D Environment" (ICRA 2011). In this implementation, a `motion_target` is set at a distance away from the robot that is exponentially stable to generate a smooth trajectory for the robot to follow.
+
+See the package's ``README`` for more complete information.
+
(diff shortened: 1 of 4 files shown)


configuration/packages/configuring-graceful-motion-controller.rst
```diff
@@ -0,0 +1,232 @@
...
+  ============== =============================
+  Type           Default                      
+  -------------- -----------------------------
+  double         0.75 
+  ============== =============================
+
+  Description
+    The difference in the path orientation and the starting robot orientation to trigger a rotate in place, if ``initial_rotation`` is enabled.
+
+:initial_rotation:
```
SteveMacenski, line 160: Appears twice? Should be final rotation?

SteveMacenski (changes requested):

configuration/packages/configuring-graceful-motion-controller.rst
ajtudela: Fixed now!

SteveMacenski:
@ajtudela thanks for the contribution! If you had a video of this controller in action, that would be great for an announcement post to the community about your new controller plugin! Its worth spending a few minutes to make a nice demo video (then, that could be added to the config guide page as an example!)

ajtudela:
I totally forgot! I made this [video](https://youtu.be/P-iKCo4h56M)

I don't know if it worth it.

SteveMacenski:
That video is a bit slow and glitchy (lots of jumps), if you could make another one at more reasonable speeds and smoothness, that would be great!

ajtudela:
Hi, I didn't forget about the video. It seems that YTdecrese the bitrate too much. So, I'd try to make a new video during this week.

SteveMacenski:
It seems a bit wobbly when there are small deviations in the path that the control point jitters between. Is there a nice way to smooth that out? (@mikeferguson perhaps a nice challenge when you're working on the dynamic point setting)

It definitely seems like keeping the point further off is a good move.

PR #518, opened 2024-02-06
fixing requirements

padhupradheep:
okay my try to fix the requirements and conflicts.. 

#364

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,5 @@
 breathe
-docutils==0.16
+docutils
 jinja2==3.0.3
 myst-parser
 sphinx_copybutton==0.5.2
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,5 +5,5 @@ myst-parser
 sphinx_copybutton==0.5.2
 sphinx_rtd_theme
 sphinx-autobuild
-sphinx==4.2.0
+sphinx
 sphinxcontrib-plantuml
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,9 @@
-breathe==4.28.0
-docutils==0.14
+breathe
+docutils==0.16
 jinja2==3.0.3
-myst-parser==1.0.0
+myst-parser
 sphinx_copybutton==0.5.2
 sphinx_rtd_theme
 sphinx-autobuild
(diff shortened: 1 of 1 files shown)


padhupradheep:
https://github.com/ros-planning/navigation.ros.org/pull/465#issuecomment-1721344535

I have read the comments from Alexey here. However, I cannot do such comparison since with the default setup on the main, I was able to successfully neither install nor build the packages.

SteveMacenski:
It appears to break the collapsing section in the build and install guide (https://output.circle-artifacts.com/output/job/b20b2946-bd6e-485a-91ee-3b16dd2ab84e/artifacts/0/html/development_guides/build_docs/index.html) Here live: https://navigation.ros.org/development_guides/build_docs/index.html that you can compare it to.

I suspect there are more breakages. This is why we have statically set versions to make sure new versions don't come about and break things without our note. I don't mind updating, but those updates still need to be statically set and tested to make sure there aren't any regressions (like what I mention above) or the website is updated appropriately to work around those issues. 

I understand that these are way out of date, but you should still be able to install them by setting the version number (that's what I do). With that said, I _definitely_ won't argue with someone updating them :-) However, it needs (1) new static versions to use and (2) testing by looking through the pages carefully and finding these kinds of errors and resolving them

padhupradheep:
Oh right, I forgot about the artifacts.

Could we leave this PR open? I'd try to revisit and figure out a static version for each of the dependencies when I get some time.

SteveMacenski:
OK! That should be easy though looking at the dependency install stage - does that not simply tell you what version it installed (and you could just use that)? That seems like the EZ-PZ way of doing it. If the only issue is the collapsing section, I'd be OK with just removing those as collapsing sections or having you take a look to see what else we could do to organize those long pages through another mechanism if that one can't get fixed.

I think Alexey also tried to upgrade to 22.04 instead of 20.04, which seems wise at the same time. He highlighted some of his integration changes -- though if you could tell me the difference between a few of those except Kerning (which is fine to change between versions), that would be news to me :laughing: 

So, you may be able to do even more updates to keep in line to make easier

SteveMacenski:
Has been recently resolved!

PR #520, opened 2024-02-07
update sonoma-raceway link in gps tutorial

ARK3r:
[Current link](https://navigation.ros.org/tutorials/docs/navigation2_with_gps.html#setup-gazebo-world) is broken for sure, not sure the new one is the correct one though considering it has a version in it, though it is from the stable branch.


SteveMacenski (approved):
Great, thanks!

PR #523, opened 2024-02-08
Improve description of goal_checker_plugins

ssr-yuki:
This PR complements #302, reflecting https://github.com/ros-planning/navigation2/pull/2269 in the documents.

This PR consists of ...

- modification of the type of `goal_checker_plugins` parameter and its example
- description of how to handle `goal_checker_id` field in `FollowPath` actions with the parameter

### DIFF
--- a/configuration/packages/configuring-controller-server.rst
+++ b/configuration/packages/configuring-controller-server.rst
@@ -102,14 +102,14 @@ Parameters
 
 :goal_checker_plugins:
 
-  ============== ==============
+  ============== ================
   Type           Default
-  -------------- --------------
-  string         'goal_checker'
-  ============== ==============
+  -------------- ----------------
+  vector<string> ["goal_checker"]
+  ============== ================
 
   Description
-    Mapped name for goal checker plugin for checking goal is reached.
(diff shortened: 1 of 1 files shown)


configuration/packages/configuring-controller-server.rst
```diff
@@ -105,11 +105,11 @@ Parameters
   ============== ==============
   Type           Default
   -------------- --------------
-  string         'goal_checker'
+  vector<string> ["goal_checker"]
   ============== ==============
```
SteveMacenski, line 109: Its pedantic, but the CI job complains: the table's top/bottom/dividers need to be the same length or larger than the content. This overflows in the 'value' part by 2 chars

SteveMacenski (changes requested):

configuration/packages/configuring-controller-server.rst
ssr-yuki: I've fixed this :+1:

SteveMacenski (approved):
Thanks! I try very hard to keep the docs in line with main, but sometimes something slips through the cracks :(

PR #524, opened 2024-02-09
Update the default ports

facontidavide:
Changes in the documentationrelated to:

Related to https://github.com/ros-planning/navigation2/pull/4060


configuration/packages/bt-plugins/conditions/GloballyUpdatedGoal.rst
```diff
@@ -9,6 +9,31 @@ Returns failure if the goal is the same, if it changes, it returns success.
 This node differs from the GoalUpdated by retaining the state of the current goal/goals throughout each tick of the BehaviorTree
 such that it will update on any "global" change to the goal. 
 
+Input Ports
+-----------
+
+:goal:
+
+  =============================== =======
```
SteveMacenski, line 17: For all: the `rst` format needs the table contents to fit in the table bounds. If you add 1 more char for the `=` and `-` lines, this is good to go.

SteveMacenski (changes requested):

PR #548, opened 2024-04-25
Replace incorrect 'arguments' with 'parameters'

kjrusscher:
Launching this code with the option 'gui:=False' gives errors in RVIZ. The correct way of using parameters in the code is 'parameters' argument. The argument 'arguments' is for command line arguments.

### DIFF
--- a/setup_guides/urdf/setup_urdf.rst
+++ b/setup_guides/urdf/setup_urdf.rst
@@ -220,7 +220,7 @@ Next, let us create our launch file. Launch files are used by ROS 2 to bring up 
           package='joint_state_publisher',
           executable='joint_state_publisher',
           name='joint_state_publisher',
-          arguments=[default_model_path],
+          parameters=[default_model_path],
           condition=launch.conditions.UnlessCondition(LaunchConfiguration('gui'))
       )
       joint_state_publisher_gui_node = launch_ros.actions.Node(


setup_guides/urdf/setup_urdf.rst
```diff
@@ -220,7 +220,7 @@ Next, let us create our launch file. Launch files are used by ROS 2 to bring up
           package='joint_state_publisher',
           executable='joint_state_publisher',
           name='joint_state_publisher',
-          arguments=[default_model_path],
+          parameters=[default_model_path],
```
SteveMacenski, line 223: I believe parameters needs to have path set to the appropriate parameter name, no? This is just a filepath

kjrusscher: I guess that the code infers the parameter name from the file extension or something like that. For example, in the "Setting up Odometry" tutorial is this code:
```
robot_localization_node = launch_ros.actions.Node(
       package='robot_localization',
       executable='ekf_node',
       name='ekf_filter_node',
       output='screen',
       parameters=[os.path.join(pkg_share, 'config/ekf.yaml'), {'use_sim_time': LaunchConfiguration('use_sim_time')}]
)
```
Here, parameters is a list that also contains a filepath.

But I don't have a lot of experience with ros2 launch files. I've tested the launch file with `parameters=[default_model_path],` and can confirm that it works. The code with `arguments=[default_model_path],` definitely does not work.

SteveMacenski: Those are yaml files of parameters and are a supported way of handling that. This is a URDF file that I assume should be fed into a specific parameter?

kjrusscher: I've improved the code, see #550

kjrusscher:
New pull request #550

PR #566, opened 2024-06-11
update behaviour tree version

ARK3r:
Looking at [a behavior tree example from nav2_bt_navigator](https://github.com/ros-navigation/navigation2/blob/main/nav2_bt_navigator/behavior_trees/navigate_to_pose_w_replanning_and_recovery.xml#L7) I see the version was changed 4 months ago but the doc wasn't updated.


SteveMacenski (approved):

SteveMacenski:
Thanks!

PR #600, opened 2024-10-18
Add missing params for loopback_sim

adivardi:


configuration/packages/configuring-loopback-sim.rst
```diff
@@ -61,13 +61,57 @@ Parameters
...
   Description
-    The scan frame to use to publish a scan for collision monitor's happiness
+    The scan frame to use to publish a scan
+
+:enable_stamped_cmd_vel:
+
+  ============== ==============
+  Type           Default
+  -------------- --------------
+  string         `false`
```
SteveMacenski, line 77: Remove the "`" to be inline with other param tables. 

Also, for rst format, you would have had to use 2x of those to get it to consider a code block. 1x is in italics

SteveMacenski (changes requested):
Otherwise, happy to merge

configuration/packages/configuring-loopback-sim.rst
adivardi: fixed both issues

SteveMacenski (approved):
Sorry for the delay, we had ROSCon, ROS-I, and then I took a short vacation while in Europe. I'm getting back to work this morning :-)

PR #604, opened 2024-11-12
DR Formatting edits

drobb257:


drobb257:
Pull request created in error, meant to create a pull request on a fork

SteveMacenski:
OK. Is the intent to open a PR on Nav2's docs for this? If so, we should probably chat (steve@opennav.org)

PR #609, opened 2024-11-20
Add note on multirobot refactor to migration guide

luca-della-vedova:
Add migration for PR https://github.com/ros-navigation/navigation2/pull/4715

### DIFF
--- a/configuration/packages/costmap-plugins/obstacle.rst
+++ b/configuration/packages/costmap-plugins/obstacle.rst
@@ -88,6 +88,7 @@ This costmap layer implements a plugin that uses 2D raycasting for 2D lidars, de
 
   Relative topics will be relative to the node's parent namespace.
   For example, if you specify `topic: scan` in the `obstacle_layer` of a `local_costmap` and you launch your bringup with a `tb4` namespace:
+
   * User chosen namespace is `tb4`.
   * User chosen topic is `scan`.
   * Topic will be remapped to `/tb4/scan` without `local_costmap`.
--- a/configuration/packages/costmap-plugins/obstacle.rst
+++ b/configuration/packages/costmap-plugins/obstacle.rst
@@ -86,6 +86,13 @@ This costmap layer implements a plugin that uses 2D raycasting for 2D lidars, de
   Description
     Topic of data.
 
+  Relative topics will be relative to the node's parent namespace.
+  For example, if you specify `topic: scan` in the `obstacle_layer` of a `local_costmap` and you launch your bringup with a `tb4` namespace:
+  * User chosen namespace is `tb4`.
+  * User chosen topic is `scan`.
+  * Topic will be remapped to `/tb4/scan` without `local_costmap`.
+  * Use global topic `/scan` if you do not wish the node namespace to apply.
+
 :``<obstacle layer>``. ``<data source>``.sensor_frame:
 
   ====== =======
--- a/configuration/packages/costmap-plugins/range.rst
+++ b/configuration/packages/costmap-plugins/range.rst
@@ -31,6 +31,7 @@ This costmap layer implements a plugin that processes sonar, IR, or other 1-D se
 
   Relative topics will be relative to the node's parent namespace.
   For example, if you specify `topics: [range1, /range2]` in the `range_layer` of a `local_costmap` and you launch your bringup with a `tb4` namespace:
+
   * User chosen namespace is `tb4`.
   * User chosen topics are [`range1`, `/range2`].
   * Topic will be remapped to `/tb4/range1`, without `local_costmap`, and `/range2`.
--- a/configuration/packages/costmap-plugins/range.rst
+++ b/configuration/packages/costmap-plugins/range.rst
@@ -29,6 +29,13 @@ This costmap layer implements a plugin that processes sonar, IR, or other 1-D se
   Description
     Range topics to subscribe to.
 
+  Relative topics will be relative to the node's parent namespace.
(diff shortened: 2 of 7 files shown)


migration/Jazzy.rst
```diff
@@ -99,3 +99,16 @@ In `PR #4675 <https://github.com/ros-navigation/navigation2/pull/4675>`_ a ``pub
 Default value:
 
 - true
+
+Revamped multirobot bringup and config files to use namespaces
+*******************************************************
```
SteveMacenski, line 104: The `***`'s annoyingly need to match the length of the title in rst in order to render properly. 

Also, for those sensor data topics that were updated with the namespacing changes, please add a note like the inline code example you provided in the obstacle layer for how to configure it (and recommend `/` prefix if you dont want to deal with namespaces). That way, its clearly up-front for users reading the docs the expected behavior they'll see without guessing and checking

SteveMacenski (changes requested):

migration/Jazzy.rst
luca-della-vedova: https://github.com/ros-navigation/docs.nav2.org/pull/609/commits/03da77955501e786e1734cd8c80626dfe5727068

SteveMacenski: Please also add that note in the `topic` configuration guide entries i.e. 
- https://docs.nav2.org/configuration/packages/costmap-plugins/obstacle.html
- https://docs.nav2.org/configuration/packages/costmap-plugins/voxel.html
- https://docs.nav2.org/configuration/packages/costmap-plugins/range.html
- https://docs.nav2.org/configuration/packages/costmap-plugins/static.html

I want this to be really loud and clear

luca-della-vedova: https://github.com/ros-navigation/docs.nav2.org/pull/609/commits/c42eb570e33d608e7d3154c96de94e4f19a2300a

configuration/packages/costmap-plugins/obstacle.rst
```diff
@@ -86,6 +86,13 @@ This costmap layer implements a plugin that uses 2D raycasting for 2D lidars, de
   Description
     Topic of data.
 
+  Relative topics will be relative to the node's parent namespace.
```
SteveMacenski, line 89: These don't render properly: https://output.circle-artifacts.com/output/job/190ba9cc-e4b2-4590-b0e4-838801870a4f/artifacts/0/html/configuration/packages/costmap-plugins/obstacle.html

I think that's fixed by adding an empty line before the first bullet (after the line ending with `:` to separate it)

SteveMacenski:
One more thing and this should be done https://docs.nav2.org/tuning/index.html --> remove the `use_namespace` parameter description if we're removing it

SteveMacenski (approved):

SteveMacenski:
I fixed some formatting for you, LGTM

luca-della-vedova:
It shows that I only wrote markdown in my life :sweat_smile:, noticed that the highlighting of snippets wasn't working because it's supposed to be double backtick so fixed that in https://github.com/ros-navigation/docs.nav2.org/pull/609/commits/3cc1e2e5cbde230740064556c9a450d0d6238a69

configuration/packages/costmap-plugins/range.rst
```diff
@@ -29,6 +29,14 @@ This costmap layer implements a plugin that processes sonar, IR, or other 1-D se
   Description
     Range topics to subscribe to.
 
+  Relative topics will be relative to the node's parent namespace.
```
SteveMacenski, line 32: Missed the entry for `map_topic` now  :wink: 

Sorry to be a stickler about docs, I know its frustrating :(

SteveMacenski (changes requested):

SteveMacenski:
Otherwise, I can merge

configuration/packages/costmap-plugins/range.rst
luca-della-vedova: Apologies not sure I follow, I thought `map_topic` is only a parameter in `static_layer`? Why is this comment on `range_sensor_layer`?

```
lucadv@noble:~/ionic_ws/src/nav2/navigation2/nav2_costmap_2d/plugins$ grep -r . -e map_topic
./static_layer.cpp:    map_topic_.c_str(),
./static_layer.cpp:    map_topic_, map_qos,
./static_layer.cpp:      map_topic_ + "_updates",
./static_layer.cpp:  declareParameter("map_topic", rclcpp::ParameterValue("map"));
./static_layer.cpp:  node->get_parameter(name_ + "." + "map_topic", map_topic_);
./static_layer.cpp:  map_topic_ = joinWithParentNamespace(map_topic_);
./static_layer.cpp:      param_name == name_ + "." + "map_topic" ||
```

SteveMacenski: I mean that i wanted you to add this block wrt clarification on the relative topic namespaces into the static layer's configuration guide for the `map_topic` now that it has the same behavior profile as `topic` for range / obstacle / voxel layers

luca-della-vedova: Is https://github.com/ros-navigation/docs.nav2.org/blob/02509d147419e3d01b7a8a14d704857f2452c57a/configuration/packages/costmap-plugins/static.rst?plain=1#L65-L82 OK?

SteveMacenski: I think `Map topic to subscribe to. If left empty the map topic will default to the global `map_topic` parameter in `costmap_2d_ros`.` should be removed. The default should also be updated to `"map"` from the change as well. Otherwise, LGTM

SteveMacenski (approved):

PR #611, opened 2024-11-22
fix typo in BT node plugin list

ottojo:
Fix typo in link target for the "Reinitialize Global Localization Service" BT node


SteveMacenski (approved):
Thanks!

PR #612, opened 2024-11-27
Add docs for IsStopped BT node

tonynajjar:
https://github.com/ros-navigation/navigation2/pull/4764


SteveMacenski:
@tonynajjar I merged too fast, it wasn't added to an index for it to be visible on the explorer https://github.com/ros-navigation/docs.nav2.org/blob/master/configuration/packages/configuring-bt-xml.rst

tonynajjar:
I accidentally already pushed the fix to master without a PR, I hope the fix is okay

SteveMacenski:
Not ideal process, but the solution LGTM

PR #615, opened 2024-12-03
Plugin container layer docs

alexanderjyuen:
Added documentation for plugin container layer


configuration/packages/costmap-plugins/plugin_container.rst
```diff
@@ -0,0 +1,34 @@
...
+  -------------- -------
+  vector<string> {}
+  ============== =======
+
+  Description
+    List of mapped costmap layer names for parameter namespaces and names.
+
+  Note
+    Costmap filters are presently unsupported
+
```
SteveMacenski, line 34: At the end here is where I'd recommend having the example YAML configuration. You could also get away without doing a separate tutorial if you included an image of that example YAML with 2x of these containers surrounding 2x dfiferent layers with different inflations to show case it "working" to prove the point visually 

This is also missing a parameter for the plugin types (the `nav2_util` plugin loader has one internal, its `<plugin_name>.plugins`)

plugins/index.rst
```diff
@@ -70,6 +70,10 @@ Costmap Layers
 |                                |                        | standalone obstacles or small    |
 |                                |                        | obstacles groups                 |
 +--------------------------------+------------------------+----------------------------------+
+| `Plugin Container Layer`_      |  Alexander Yuen        | Combines the different layers    |
```
SteveMacenski, line 73: Maybe reword this to be a little more clear :upside_down_face:

tuning/index.rst
```diff
@@ -133,6 +133,7 @@ Costmap2D has a number of plugins that you can use (including the availability f
 - ``SpatioTemporalVoxelLayer``: 3D costmap layer for 3D lidars, non-planar 2D lidars, or depth camera processing based on temporal decay. Useful for robots with high sensor coverage like 3D lidars or many depth cameras at a reduced computational overhead due to lack of raycasting.
 - ``RangeLayer``: Models sonars, IR sensors, or other range sensors for costmap inclusion
 - ``DenoiseLayer``: Removes salt and pepper noise from final costmap in order to remove unfiltered noise. Also has the option to remove clusters of configurable size to remove effects of dynamic obstacles without temporal decay.
+- ``PluginContainerLayer``: Combines the layers above to allow for different submaps to exist in the map. This would allow applying different denoise layers or inflation layers to different obstacle or static layers in the same costmap
```
SteveMacenski, line 136: Ditto

SteveMacenski (changes requested):
Just add to the migration guide that "hey we have a new layer, here's why its nice, check out the configuration guide" with a link. Its a promotion :-)

configuration/packages/costmap-plugins/plugin_container.rst
alexanderjyuen: Is this necessary? I am presently getting the `plugin_types` in the [following way](https://github.com/alexanderjyuen/navigation2/blob/feature-plugin-container-layer/nav2_costmap_2d/plugins/plugin_container_layer.cpp#L69)

alexanderjyuen:
Example image updated

configuration/packages/costmap-plugins/plugin_container.rst
alexanderjyuen: Added to the migration guide

configuration/packages/costmap-plugins/plugin_container.rst
```diff
@@ -0,0 +1,95 @@
...
+      plugins: ["plugin_container_static_layer", "plugin_container_obstacle_layer"]
+
+      plugin_container_static_layer:
+        plugin: "nav2_costmap_2d::PluginContainerLayer"
+        enabled: True
+        combination_method: 1
+        plugins: ["static_layer", "inflation_layer"]
+        static_layer:
+          plugin: "nav2_costmap_2d::StaticLayer"
```
alexanderjyuen, line 57: @SteveMacenski if the plugin type is explicitly defined in each constituent layer, do I still need an parameter for `plugin_types`

SteveMacenski: Its not defined in the layer as I mentioned. Your util that loads the plugin (https://github.com/ros-navigation/navigation2/blob/684a4d037241674550b756c5014711fa6c13e35e/nav2_util/include/nav2_util/node_utils.hpp#L136) itself defines that `.plugin` parameter so it can read the string of the plugin class in order to load it. That is a parameter that you have actually added by using `get_plugin_type_param`

SteveMacenski:
Otherwise, with that missing parameter + fixing DCO, I can merge this and the PR!

SteveMacenski (approved):

PR #619, opened 2024-12-11
Update writing_new_costmap2d_plugin.rst

AjayAdit:
Order of plugins in plugins list has significance. This was one of the insights I got from spending considerable time troubleshooting inflation error I got when I added range layer. Hoping henceforth this piece of insight will help new user. Please do let me know if this is explicitly mentioned elsewhere in the documentation.


SteveMacenski (approved):
Sure, sounds good!

PR #621, opened 2024-12-11
Update GoalUpdater docs

tonynajjar:
https://github.com/ros-navigation/navigation2/pull/4771

### DIFF
--- a/migration/Jazzy.rst
+++ b/migration/Jazzy.rst
@@ -51,14 +51,18 @@ Here we can see the working demo of the plugin:
 
 .. attention:: If the docking server is unavailable, then the combo box of the dock type will be empty.
 
-New BT Nodes
-************
+BT Nodes Changes
+****************
 
 Below is a list of new BT Nodes added:
 
 - ``GetPoseFromPath``: An action to get a particular pose from an input path.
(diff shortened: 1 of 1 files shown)


configuration/packages/bt-plugins/decorators/GoalUpdater.rst
```diff
@@ -19,6 +19,17 @@ Parameters
   Description
       The topic to receive the updated goal pose
 
+:goals_updater_topic:
+
+  ====== ==============
+  Type   Default
+  ------ --------------
+  string  "goals_update"
+  ====== ==============
```
SteveMacenski, line 28: the table needs to be expanded by 1 `=` and `-` since the content bleeds outside of it 

... rst is really annoying

SteveMacenski (changes requested):
Otherwise, :+1:

SteveMacenski (approved):
Please add the migration guide too. You can add it under the `New BT Nodes[](https://docs.nav2.org/migration/Jazzy.html#new-bt-nodes)` section I think (with a small title rename to BT Node Changes)

SteveMacenski:
@tonynajjar just add to the migration guide and we should be Good to Go to merge this too

tonynajjar:
Done

PR #624, opened 2024-12-13
update docs for iterative graceful controller

mikeferguson:
Doc changes for https://github.com/ros-navigation/navigation2/pull/4795


mikeferguson:
TODO (after review and updates to underlying PR):
 * [ ] re-add `initial_rotation`
 * [ ] remove `add_orientations`
 * [ ] document `in_place_collision_tolerance`
 * [ ] Maybe improve discussion of what `prefer_final_rotation` actually means

SteveMacenski (approved):
Approved pending underlying PR.

Only thing would be to add in the migration guide that we adjusted some parameters and list just the changed parameters for folks to be aware of

SteveMacenski (commented):
@mikeferguson a few additions:
- You changed the defaults of the k/beta parameters, the new defaults should be updated in this table
- In the migration guide, we should mention this (Pretty big) change 
- In the Navigation Plugins table, I'd add in omni as supported motion models (not just diff)

mikeferguson:
> * In the migration guide, we should mention this (Pretty big) change

Where in the migration doc do we want this? The Jazzy->K-turtle part, or are we planning to backport at all?

SteveMacenski:
Since it changes and removes parameters, it cannot be backported, so J->K

mikeferguson:
@SteveMacenski - added requested items - I think this ready for a final review

SteveMacenski (approved):

PR #630, opened 2025-01-07
PoseStampedArray docs

tonynajjar:
Docs for https://github.com/ros-navigation/navigation2/pull/4791


migration/Jazzy.rst
```diff
@@ -1,9 +1,9 @@
 .. _jazzy_migration:
 
-Jazzy to K-Turtle
+Jazzy to Kilted
 #################
```
SteveMacenski, line 4:
```suggestion
###############
```

Lengths need to match, annoyingly with rst

migration/Jazzy.rst
```diff
@@ -214,3 +214,10 @@ In `PR #4795 <https://github.com/ros-navigation/navigation2/pull/4795>`_ the ``n
 * Automatic creation of orientations for the plan if they are missing.
 * Addition of ``v_angular_min_in_place`` parameter to avoid the robot getting stuck while rotating due to mechanical limitations.
 * ``final_rotation`` has been renamed ``prefer_final_rotation`` and the behavior has changed slightly.
+
+Introduction of PoseStampedArray
```
SteveMacenski, line 218: I'd retitle this that the ComputePathThroughPoses / NavigateThroughPoses now relies on PoseStampedArray instead. It puts the details on "what changed" versus the implementation details of the change

migration/Jazzy.rst
tonynajjar: I will never get this right :facepalm:

tonynajjar:
> Does any of the docs w.r.t. Nav2 BT explanations and walk throughs need to be updated?

As far as I see not

SteveMacenski (approved):

PR #631, opened 2025-01-07
Add note about multi-robot launch file changes

TannerGilbert:
Documentation changes for https://github.com/ros-navigation/navigation2/pull/4811.

Should I add an example of what was the problem with the original parsing namely that something like the following didn't work as expected:
```python
IncludeLaunchDescription(
    PythonLaunchDescriptionSource(
        os.path.join(get_package_share_directory('nav2_bringup'), "launch", "cloned_multi_tb3_simulation_launch.py")
    ),
    launch_arguments={"robots": "robot1={x: 0.5, y: 0.5, yaw: 1.5707}"}.items(),
)
```

or is this not relevant for the documentation?

### DIFF
--- a/migration/Jazzy.rst
+++ b/migration/Jazzy.rst
@@ -214,3 +214,8 @@ In `PR #4795 <https://github.com/ros-navigation/navigation2/pull/4795>`_ the ``n
 * Automatic creation of orientations for the plan if they are missing.
 * Addition of ``v_angular_min_in_place`` parameter to avoid the robot getting stuck while rotating due to mechanical limitations.
 * ``final_rotation`` has been renamed ``prefer_final_rotation`` and the behavior has changed slightly.
+
+Conform to ROS 2 launch syntax in Turtlebot 3 multi-robot launch file
+*********************************************************************
+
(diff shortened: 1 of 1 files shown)


migration/Jazzy.rst
```diff
@@ -214,3 +214,8 @@ In `PR #4795 <https://github.com/ros-navigation/navigation2/pull/4795>`_ the ``n
 * Automatic creation of orientations for the plan if they are missing.
 * Addition of ``v_angular_min_in_place`` parameter to avoid the robot getting stuck while rotating due to mechanical limitations.
 * ``final_rotation`` has been renamed ``prefer_final_rotation`` and the behavior has changed slightly.
+
+Conform to ROS 2 launch syntax in Turtlebot 3 multi-robot launch file
+*********************************************************************
+
+In `PR #4811 <https://github.com/ros-navigation/navigation2/pull/4811>`_ the ``cloned_multi_tb3_simulation_launch.py`` launch file was updated so that parsing the robots conforms to the ROS 2 launch file standards. This was achieved by refactoring the ``ParseMultiRobotPose`` class to be a custom launch substitution. This change allows users to pass the ``robots`` from another launch file through ``launch_arguments`` which was not possible with the old version.
```
SteveMacenski, line 221: Can you include the blob of what you did with `launch_arguments` so that someone else has an easier time doing it themselves :-)

SteveMacenski: You can do it via 

```
.. code-block:: python

   my_python_with_a_space_after_code_block_and_indented_2_spaces()

```

SteveMacenski (approved):
Just waiting on CI so I can check the output

PR #633, opened 2025-01-14
Tutorial for Nav2 Route Tool

john-chrosniak:
This PR includes a tutorial for the Nav2 route tool that was introduced here: https://github.com/ros-navigation/navigation2/pull/4775


tutorials/docs/navigation2_route_tool.rst
```diff
@@ -0,0 +1,99 @@
...
+The Nav2 route tool is an Rviz panel designed to ease the process of creating and editing the route graphs used by the Nav2 Route Server.
+In contrast to a free-space planner, route graphs provided a structure path for mobile robots to follow when navigating through their environment.
+Supported operations for the route tool include loading in existing route graphs, adding new nodes and edges, editing existing nodes and edges, removing nodes and edges, and saving changes to a new file.
+By the end of this tutorial, you should feel comfortable creating and modifying route graphs using the route tool.
+
+Prerequisites
+============
+
+Prior to using the route tool, you will need to create an occupancy grid map using the slam_toolbox package. 
+Follow `this tutorial <https://roboticsbackend.com/ros2-nav2-generate-a-map-with-slam_toolbox/>`_ for instructions on how to generate and save an occupancy grid map.
```
SteveMacenski, line 26: Remove the external tutorial, I've had bad luck with websites goign down / deleting pages that leave dead links. I think https://docs.nav2.org/tutorials/docs/navigation2_with_slam.html does a good job of this instead so the website is closed-loop

SteveMacenski (changes requested):
I think there's some git issues since it looks like there are unrelated commits to this in the diff (ie updates to the graceful controller config page, costmap config page, etc) that needs to be sorted :-) 

These gifs are _fantastic_! They are really tutorials in it of themselves, so I really appreciate that attention to detail and quality. 

The only other thing that might be nice is then using that graph to do a planning request with the route server, but maybe that's better served in a new tutorial (out of scope of your needs?). What do you think?

Overall, this looks great and I don't have really many notes

tutorials/docs/navigation2_route_tool.rst
john-chrosniak: Got it, just replaced this link

john-chrosniak:
> I think there's some git issues since it looks like there are unrelated commits to this in the diff (ie updates to the graceful controller config page, costmap config page, etc) that needs to be sorted :-)

Should be resolved now!

> The only other thing that might be nice is then using that graph to do a planning request with the route server, but maybe that's better served in a new tutorial (out of scope of your needs?). What do you think?

I think this may be out of scope for the route generation tool tutorial but I'd be happy to create this as part of a separate route server tutorial. Appreciate the feedback!

SteveMacenski (approved):
I'll merge this in once we have the route server work in Nav2, but it is approved and will be merged in as soon as the tool is available as well :-)

SteveMacenski:
> I think this may be out of scope for the route generation tool tutorial but I'd be happy to create this as part of a separate route server tutorial. Appreciate the feedback!

Upon further thinking, I think this is something I should put together. I have some subtle details I want to explain about how to use the route server (i.e. use for global routing -> then a global planner for a local horizon -> then trajectory generator VS global routing and using this as the plan for trajectory generation VS global routing and as global plan being tracked using the tracking capabilities and replanning within the route server). I think this would need to wait until that's all done and I can put demos together since some of those things aren't 100% complete and would also require building a few BT demos of nodes that also don't exist yet

SteveMacenski:
Toggling to retrigger CI so I can get the build artifacts back -- this is on my radar for next week as I'm making headway into the updates!

PR #637, opened 2025-01-28
Add new parameter description in ispathvalid

mini-1235:
https://github.com/ros-navigation/navigation2/pull/4873

### DIFF
--- a/configuration/packages/bt-plugins/conditions/IsPathValid.rst
+++ b/configuration/packages/bt-plugins/conditions/IsPathValid.rst
@@ -5,8 +5,7 @@ IsPathValid
 
 Checks to see if the global path is valid. If there is a
 obstacle along the path, the condition returns FAILURE, otherwise
-it returns SUCCESS. Additionally, users can specify a threshold cost(0-254)
-for path validity. If the path cost exceeds this threshold, the condition will also return FAILURE.
+it returns SUCCESS.
 
 Input Ports
 -----------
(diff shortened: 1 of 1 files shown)


configuration/packages/bt-plugins/conditions/IsPathValid.rst
```diff
@@ -5,11 +5,12 @@ IsPathValid
 
 Checks to see if the global path is valid. If there is a
 obstacle along the path, the condition returns FAILURE, otherwise
-it returns SUCCESS. 
+it returns SUCCESS. Additionally, users can specify a threshold cost(0-254)
+for path validity. If the path cost exceeds this threshold, the condition will also return FAILURE.
```
SteveMacenski, line 9: Please put in the new port field - see https://github.com/ros-navigation/docs.nav2.org/blob/c3a899333e06baaabdc3ede6f044a403c2d7b3b7/configuration/packages/bt-plugins/conditions/GoalReached.rst for example for the table entries format for the new `max_cost` port (default 253)

SteveMacenski (changes requested):

configuration/packages/bt-plugins/conditions/IsPathValid.rst
mini-1235: Hi, do I need to do it for `server_timeout` as well?

mini-1235: > Hi, do I need to do it for `server_timeout` as well?

I saw it in the document of some action nodes, but I didn't see it on the original ispathvalid document

SteveMacenski: Yeah, please add that + the path, that seems to be an oversight on my part when this PR was merged that I didn't enforce the full table. I'd appreciate if you filled in the gaps from my previous reviews :laughing:

SteveMacenski (approved):

PR #639, opened 2025-01-30
Rotation Shim closed_loop param

RBT22:
Add docs for https://github.com/ros-navigation/navigation2/pull/4880

### DIFF
--- a/configuration/packages/configuring-rotation-shim-controller.rst
+++ b/configuration/packages/configuring-rotation-shim-controller.rst
@@ -138,7 +138,7 @@ Rotation Shim Controller Parameters
   ============== =============================
 
   Description
-    If false, the rotationShimController will use the last commanded velocity as the next iteration's current velocity. When acceleration limits are set appropriately, this is a good assumption. If true, it will use odometry to estimate the robot's current speed. In this case it is important that the source is high-rate and low-latency.
+    If false, the rotationShimController will use the last commanded velocity as the next iteration's current velocity. When acceleration limits are set appropriately and the robot's controllers are responsive, this can be a good assumption. If true, it will use odometry to estimate the robot's current speed. In this case it is important that the source is high-rate and low-latency to account for control delay.
 
 Example
 *******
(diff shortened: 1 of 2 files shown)


configuration/packages/configuring-rotation-shim-controller.rst
```diff
@@ -129,6 +129,17 @@ Rotation Shim Controller Parameters
...
+:closed_loop:
+
+  ============== =============================
+  Type           Default
+  -------------- -----------------------------
+  bool           true
+  ============== =============================
+
+  Description
+    If false, the rotationShimController will use the last commanded velocity as the next iteration's current velocity. When acceleration limits are set appropriately, this is a good assumption. If true, it will use odometry to estimate the robot's current speed. In this case it is important that the source is high-rate and low-latency.
```
SteveMacenski, line 141:
```suggestion
    If false, the rotationShimController will use the last commanded velocity as the next iteration's current velocity. When acceleration limits are set appropriately and the robot's controllers are responsive, this can be a good assumption. If true, it will use odometry to estimate the robot's current speed. In this case it is important that the source is high-rate and low-latency to account for control delay.
```

migration/Jazzy.rst
```diff
@@ -283,3 +283,12 @@ A demo can be seen below with the following parameters:
   :width: 800
   :alt: Rotation Shim Deceleration
   :align: center
+
+Rotation Shim Open-loop Control
+*******************************
+
+The parameter ``closed_loop`` was introduced to the Rotation Shim Controller to allow users to choose between open-loop and closed-loop control. When set to ``false``, the controller will use the last commanded velocity as the next iteration's current velocity. When acceleration limits are set appropriately, this is a good assumption and doesn't require high-rate, low-latency odometry.
```
SteveMacenski, line 290: Update here for above

SteveMacenski (commented):
LGTM

SteveMacenski (approved):

PR #640, opened 2025-02-01
Added restore_cleared_footprint parameter for static costmap layer

CihatAltiparmak:


configuration/packages/costmap-plugins/static.rst
```diff
@@ -29,6 +29,17 @@ This implements a costmap layer taking in a map from either SLAM or ``map_server
   Description
     Clear any occupied cells under robot footprint.
 
+:``<static layer>``.restore_outdated_map:
```
SteveMacenski, line 32: See PR.

Also, lets default to `true`

SteveMacenski (approved):

PR #643, opened 2025-02-05
Remove use of default_bt_xml_filename

mini-1235:


SteveMacenski (approved):

SteveMacenski:
Thanks for all of these doc updates - they're great!

PR #648, opened 2025-02-21
Edit port table to stay consistent with code

mini-1235:


SteveMacenski (approved):

SteveMacenski:
thanks for all of these!

PR #650, opened 2025-02-27
Update setup_robot_localization.rst

coffinfish:
Replaced instances of ekf_filter_node with ekf_node

The commands refer to the node as ekf_filter_node when it should be ekf_node (as it is named in the launch file and ekf.yaml file). Apologies if this is incorrect/unintended.


SteveMacenski:
@coffinfish I think actually the code should be fixed to use `ekf_filter_node` so that it is consistent with the other distribution branches. We have tutorials for pre-Jazzy using `ekf_filter_node` and switching the name seems unncessary in the code.

If you update these launch files and the name in the config file, that would have the same impact: https://github.com/search?q=repo%3Aros-navigation%2Fnavigation2_tutorials%20ekf_node&type=code

coffinfish:
That makes a lot of sense. I (hopefully) ended up doing the changes you mentioned above correctly. I kept the most of the original documentation the same and only changed the config yaml file and the name parameter in the launch file.

SteveMacenski:
@coffinfish I tihnk to complete this, you'd need to submit a PR to change the names in the `navigation2_tutorials` repository to match that name as well https://github.com/search?q=repo%3Aros-navigation%2Fnavigation2_tutorials%20ekf_node&type=code

PR #658, opened 2025-03-17
replaced <ros2-distro> with $ROS_DISTRO

glitchhopcore:
We have specified `<ros2-distro>` to refer to the ROS2 version, instead of which we can simply use $ROS_DISTRO, which would make it more copy-paste friendly.

### DIFF
--- a/getting_started/index.rst
+++ b/getting_started/index.rst
@@ -28,8 +28,8 @@ Thus, for Jazzy and newer, the installation packages and instructions are slight
 
    .. code-block:: bash
 
-      sudo apt install ros-<ros2-distro>-navigation2
-      sudo apt install ros-<ros2-distro>-nav2-bringup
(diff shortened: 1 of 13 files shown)


getting_started/index.rst
```diff
@@ -54,9 +54,9 @@ Running the Example
 
    .. code-block:: bash
 
-      source /opt/ros/<ros2-distro>/setup.bash
+      source /opt/ros/$ROS_DISTRO/setup.bash
```
SteveMacenski, line 57: Just a note: the `ROS_DISTRO` env variable isn't set before you source the setup.bash file, so that would not be in your environment yet to work. This is actually true of several of the changes in this PR

mini-1235:
It looks like https://github.com/ros-navigation/docs.nav2.org/pull/863 supersedes this

@glitchhopcore Thanks for the effort here!

PR #666, opened 2025-04-02
Add Rotation Shim Controller Use Path Orientations

bektaskemal:
See https://github.com/ros-navigation/navigation2/pull/5034

### DIFF
--- a/configuration/packages/configuring-rotation-shim-controller.rst
+++ b/configuration/packages/configuring-rotation-shim-controller.rst
@@ -140,6 +140,17 @@ Rotation Shim Controller Parameters
   Description
     If false, the rotationShimController will use the last commanded velocity as the next iteration's current velocity. When acceleration limits are set appropriately and the robot's controllers are responsive, this can be a good assumption. If true, it will use odometry to estimate the robot's current speed. In this case it is important that the source is high-rate and low-latency to account for control delay.
 
+:use_path_orientations:
+
+  ============== ===========================
+  Type           Default
+  -------------- ---------------------------
+  bool           false
+  ============== ===========================
+
+  Description
+    If true, the controller will use the orientations of the path points to compute the heading of the path instead of computing the heading from the path points. If true, the controller will use the orientations of the path points to compute the heading of the path instead of computing the heading from the path points. Use for for feasible planners like the Smac Planner which generate feasible paths with orientations for forward and reverse motion.
+
 Example
 *******
 .. code-block:: yaml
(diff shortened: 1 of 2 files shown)


migration/Jazzy.rst
```diff
@@ -362,3 +362,12 @@ In `PR #4996 <https://github.com/ros-navigation/navigation2/pull/4996>`_ a new p
 Default value:
 
 - 253
+
+Rotation Shim Using Path Orientations
+*******************************
```
SteveMacenski, line 367:
```suggestion
*************************************
```

migration/Jazzy.rst
```diff
@@ -362,3 +362,12 @@ In `PR #4996 <https://github.com/ros-navigation/navigation2/pull/4996>`_ a new p
...
+
+The parameter ``use_path_orientations`` was introduced to the Rotation Shim Controller to allow users to choose between using the orientations of the path points to compute the heading of the path instead of computing the heading from the path points. This is useful for planners that generate feasible paths with orientations for forward and reverse motion, such as the Smac Planner.
```
SteveMacenski, line 369:
```suggestion
The parameter ``use_path_orientations`` was introduced to the Rotation Shim Controller to allow users to choose between using the orientations of the path points to compute the heading of the path instead of computing the heading from the path point's relative locations. This is useful for planners that generate feasible paths with orientations for forward and reverse motion, such as the Smac Planner.
```

configuration/packages/configuring-rotation-shim-controller.rst
```diff
@@ -140,6 +140,17 @@ Rotation Shim Controller Parameters
...
+:use_path_orientations:
+
+  ============== ===========================
+  Type           Default
+  -------------- ---------------------------
+  bool           false
+  ============== ===========================
+
+  Description
+    If true, the controller will use the orientations of the path points to compute the heading of the path instead of computing the heading from the path points. If true, the controller will use the orientations of the path points to compute the heading of the path instead of computing the heading from the path points. Use for for feasible planners like the Smac Planner which generate feasible paths with orientations for forward and reverse motion.
```
SteveMacenski, line 152:
```suggestion
    If true, the controller will use the orientations of the path points to compute the heading of the path instead of computing the heading from the path point's relative locations. If true, the controller will use the orientations of the path points to compute the heading of the path instead of computing the heading from the path points. Use for for feasible planners like the Smac Planner which generate feasible paths with orientations for forward and reverse motion.
```

SteveMacenski (approved):

PR #676, opened 2025-04-11
fix: deprecated tag `<ray>`, use `<lidar>` instead

Vangasse:
While working with Gazebo Harmonic, I encountered an issue where the "ray" tag no longer seems to function as expected. After some troubleshooting, I found out by looking at  the TurtleBot3 project that the documentation is currently outdated and replacing "ray" with "lidar" resolved the problem.

### DIFF
--- a/setup_guides/sensors/setup_sensors_gz.rst
+++ b/setup_guides/sensors/setup_sensors_gz.rst
@@ -135,7 +135,7 @@ In your SDF add the below before the ``</model>`` line to add the lidar's link, 
           <update_rate>5</update_rate>
           <topic>scan</topic>
           <gz_frame_id>lidar_link</gz_frame_id>
-          <ray>
+          <lidar>
             <scan>
               <horizontal>
                 <samples>360</samples>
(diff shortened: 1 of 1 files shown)


SteveMacenski (commented):
Is there a depreciation warning? What was the issue that you ran into? 

Please sign off your commit with DCO and we can merge!

Vangasse:
Hi everyone,

Just wanted to clarify that the <ray> tag in Gazebo does still work as intended—I likely made an unrelated mistake elsewhere in the tutorial and fixed it without realizing. While <lidar> is indeed the preferred as <ray> is considered legacy, <ray> remains functional and serves the tutorial’s purpose well. As a bonus, it maintains backward compatibility with older Gazebo environments.

Since the legacy version remains fully functional, I’ll leave it to the reviewer’s discretion whether to adopt this update or maintain the current implementation. Happy to adjust either way!

Thanks for the attention.

SteveMacenski (approved):

SteveMacenski:
Thanks for the info! Merged!

PR #692, opened 2025-05-20
updated loopback sim docs

RamanRobotics:
- Updated loopback sim docs with new parameters for nav2 pull https://github.com/ros-navigation/navigation2/pull/5174


configuration/packages/configuring-loopback-sim.rst
```diff
@@ -114,6 +114,73 @@ Parameters
...
+
+  Description
+    Ending angle of the scan in radians (rightmost angle)
+
+:scan_angle_increment:
+
+  ============== ==============
+  Type           Default
+  -------------- --------------
+  double         0.0174
```
SteveMacenski, line 166: Default is `0.0261`

SteveMacenski (changes requested):

configuration/packages/configuring-loopback-sim.rst
SteveMacenski, line 166:
```suggestion
  double         0.0261
```

PR #700, opened 2025-05-23
Minor changes in Planner, Controller, Smoother, Route and Behaviour Servers in Navigation Concepts

Yasvanth-S:
Changed from Four to Five in representing action servers.

Note:
    While I was reading through the documentation I found this blunder. I don't know, if that was on purpose or really a blunder, if it is you can merge this or close this request.


SteveMacenski:
Thanks!

PR #703, opened 2025-05-28
Changed the links for Gazebo classic.

Yasvanth-S:
The current links in setting up odometry for gazebo classic is removed and ported to different domain.

I also saw two things while reading the documentation:

1. In **First Time Robot Setup Guide** page, on **Transforms in Navigation 2** topic in **Setting Up Transformations** the abbreviation for URDF is given as Universal Robot Descriptor File, I think it can also be called that way😄.
2. In the same page, on **Writing the URDF** topic in **Setting Up the URDF** there is a line  **"For code block block given below"**, when I read that first time it was confusing, are you're mentioning the code-block directive there or it's an typo.

Sorry for the long paragraphs 😄 trying to help you guys back.

### DIFF
--- a/setup_guides/odom/setup_odom_gz_classic.rst
+++ b/setup_guides/odom/setup_odom_gz_classic.rst
@@ -75,7 +75,7 @@ As an overview for this section, we will first setup Gazebo and the necessary pa
 Setup and Prerequisites
 =======================
 
-`Gazebo <http://gazebosim.org/>`_ is a 3D simulator that allows us to observe how our virtual robot will function in a simulated environment. To start using Gazebo with ROS 2, follow the installation instructions in the `Gazebo Installation Documentation <http://gazebosim.org/tutorials?cat=install>`_.
+`Gazebo <http://classic.gazebosim.org/>`_ is a 3D simulator that allows us to observe how our virtual robot will function in a simulated environment. To start using Gazebo with ROS 2, follow the installation instructions in the `Gazebo Installation Documentation <http://classic.gazebosim.org/tutorials?cat=install>`_.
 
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):

SteveMacenski:
> I think it can also be called that way

I don't understand 

>  "For code block block given below"

I also don't see that string in that page, can you elaborate?

Yasvanth-S:
> I think it can aslo be called that way

On [Transforms in Navigation2](https://docs.nav2.org/setup_guides/transformation/setup_transforms.html#transforms-in-navigation2) in the below paragraph abbreviation for <mark>URDF</mark>

>All other statically defined transforms (e.g. base_link => base_laser, base_link => wheels, wheels => IMU, etc) is what we will be talking about for the rest of this guide. This transformation tree is used by Nav2 to properly relate the information from sensors or other frame of interest to the rest of the robot. The transformation between these two coordinate frames is usually provided to Nav2 through the Robot State Publisher and the <mark>Universal Robot Descriptor File (URDF)</mark> . In cases where there are more sensor coordinate frames on your platform, then a transform tree from base_link to each sensor coordinate frame needs to be published.
---
> "For code block block given below"

On [Writing The URDF](https://docs.nav2.org/setup_guides/urdf/setup_urdf.html#writing-the-urdf)

> <mark>For the code block block below</mark>, we access the base properties from the robot constants sections we defined before using the ${property} syntax. In addition, we also set the material color of the main chassis to Cyan. Note that we set these parameters under the <visual> tag so they will only be applied as visual parameters which dont affect any collision or physical properties.

SteveMacenski:
What I meant is that I don't understand your suggested change. Should we not be expanding the abbreviation to explain what it stands for? 

> For the code block block below,

Weird, that didn't come up on my grepping before. Will fix.

PR #707, opened 2025-06-06
Update behavior tree walkthrough

mini-1235:
Hi @SteveMacenski, I learnt a lot from the "detailed behavior tree walkthrough" section, however, I noticed that the content haven't been updated for a while, I have tried to update those, can you take a look?


SteveMacenski (approved):
Thanks so much for the updates!

PR #710, opened 2025-06-13
Fix wrong config

mini-1235:


SteveMacenski:
I appreciate all these doc fixes, thanks!

PR #731, opened 2025-07-21
Add missing min_obstacle_height in obstacle layer

mini-1235:
min_obstacle_height here https://github.com/ros-navigation/navigation2/blob/27dab69edfd05f8e2c6d77d232c008a92fd5b5f5/nav2_costmap_2d/plugins/obstacle_layer.cpp#L98 is not documented 

I am also curious why does this parameter does not exist in voxel layer
https://github.com/ros-navigation/navigation2/blob/27dab69edfd05f8e2c6d77d232c008a92fd5b5f5/nav2_costmap_2d/plugins/voxel_layer.cpp#L197

### DIFF
--- a/configuration/packages/costmap-plugins/obstacle.rst
+++ b/configuration/packages/costmap-plugins/obstacle.rst
@@ -31,6 +31,17 @@ This costmap layer implements a plugin that uses 2D raycasting for 2D lidars, de
   Description
     Clear any occupied cells under robot footprint.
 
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):

SteveMacenski:
@mini-1235 does the Voxel Layer not take the obstacle layer's version of this for use? The voxel layer is derived from Obstacle. Make sure this works really quick but otherwise we should also add to the Voxel Layer's documentation

PR #757, opened 2025-08-08
Add option to override lethal cost in keepout zone

mini-1235:
https://github.com/ros-navigation/navigation2/pull/5187 missing docs


configuration/packages/costmap-plugins/keepout_filter.rst
```diff
@@ -31,6 +31,28 @@ Note: As Costmap Filters does not have the inflation layer applied to them (sinc
...
+:``<filter name>``.lethal_override_cost:
+
+  ====== =======
+  Type   Default
+  ------ -------
+  double 252
+  ====== =======
+
+  Description
+    The cost value written into those cells instead of lethal cost when override is active.
```
SteveMacenski:
```suggestion
    The cost value written into those cells instead of lethal cost when override is active. Default sets cost very high to incentivize leaving the area as soon as possible.
```

PR #794, opened 2025-10-16
Parallel state transitions

jplapp:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-navigation/navigation2/pull/5541 |
---

document parallel state transitions

### DIFF
--- a/migration/Kilted.rst
+++ b/migration/Kilted.rst
@@ -433,3 +433,10 @@ This server allows the robot to follow and maintain a determined distance from a
 The information about Following Server parameters set-up could be found at :ref:`configuring_following_server` configuration guide.
 
 The tutorial for the Following Server has been recently updated. For the latest instructions and examples, see :ref:`navigation2-dynamic-point-following`.
+
+
+
+Lifecycle Manager
+----------------
+
+`PR #5541 <https://github.com/ros-navigation/navigation2/pull/5541>`_ adds a new parallel_state_transitions parameter. This will process all state transitions in parallel which can speed up launch in systems with many lifecycle nodes. On a system with 14 lifecycle nodes, the time for activating them goes from 17 seconds to 2 seconds.


migration/Kilted.rst
```diff
@@ -433,3 +433,10 @@ This server allows the robot to follow and maintain a determined distance from a
 The information about Following Server parameters set-up could be found at :ref:`configuring_following_server` configuration guide.
 
 The tutorial for the Following Server has been recently updated. For the latest instructions and examples, see :ref:`navigation2-dynamic-point-following`.
+
+
+
+Lifecycle Manager
+----------------
```
SteveMacenski, line 440: The `--` line has to match the length of the header. Also, make this more descriptive about the specific change made

migration/Kilted.rst
```diff
@@ -433,3 +433,10 @@ This server allows the robot to follow and maintain a determined distance from a
...
+
+`PR #5541 <https://github.com/ros-navigation/navigation2/pull/5541>`_ adds a new parallel_state_transitions parameter. This will process all state transitions in parallel which can speed up launch in systems with many lifecycle nodes. On a system with 14 lifecycle nodes, the time for activating them goes from 17 seconds to 2 seconds.
```
SteveMacenski, line 442:
```suggestion
`PR #5541 <https://github.com/ros-navigation/navigation2/pull/5541>`_ adds a new parallel_state_transitions parameter. This will process all state transitions in parallel which can speed up launch in systems with many lifecycle nodes.
```

Given that's not what we see with Nav2 itself (a custom system), I'd prefer to keep that out since that makes people think Nav2 itself takes 17 seconds to load, which isn't the case I don't think?

jplapp:
closed as related PR was closed

PR #796, opened 2025-10-21
Update configuring-collision-monitor-node.rst

greganderson-vermeer:
Changes to support issue 5585 from the nav2 repo.

---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | [(Nav2 #5585](https://github.com/ros-navigation/navigation2/issues/5585l)) |
---

A use global height parameter was added to support point cloud data that contains a "height" field for calculation obstacle height. This is useful in operating domains where the ground surface is not flat or the robot is not always level to the surrounding ground level.

### DIFF
--- a/configuration/packages/collision_monitor/configuring-collision-monitor-node.rst
+++ b/configuration/packages/collision_monitor/configuring-collision-monitor-node.rst
@@ -582,7 +582,7 @@ Observation sources parameters
   ============== =============================
 
   Description:
-    Set true for pointcloud sources containing a "height" field relative to a real world ground contour. The "height" field will be used for the min and max height checks instead of the "z" field and will not be transformed with the rest of the cloud data. Applicable for ``pointcloud`` type.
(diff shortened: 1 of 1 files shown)


configuration/packages/collision_monitor/configuring-collision-monitor-node.rst
```diff
@@ -573,6 +573,17 @@ Observation sources parameters
...
+:``<source name>``.use_global_height:
+
+  ============== =============================
+  Type           Default
+  -------------- -----------------------------
+  bool           false
+  ============== =============================
+
+  Description:
+    Set true for pointcloud sources containing a "height" field relative to a real world ground contour. The "height" field will be used for the min and max height checks instead of the "z" field and will not be transformed with the rest of the cloud data. Applicable for ``pointcloud`` type.
```
SteveMacenski:
```suggestion
    Set true for pointcloud sources containing a "height" field relative to a real world ground contour. The "height" field will be used for the min and max height checks instead of the "z" field and will not be transformed as it is assumed that height is already global frame referenced. Applicable for ``pointcloud`` type.
```

PR #801, opened 2025-10-30
update installation instructions for ROS Jazzy

justagist:
## Update doc for jazzy

| ------ | ----------- |
| Ticket(s) this addresses   | update doc |
---

* Updated the documentation to make the demo work in Jazzy. I had to install `nav2-route` as well.
* Updated doc so that the command instructions can be directly copied and pasted into terminal.


justagist:
Closing this as it is not required after next jazzy sync. See https://github.com/ros-navigation/docs.nav2.org/pull/801#issuecomment-3467564539.

PR #804, opened 2025-11-03
Doc: add CostmapSource note and yaml example

Lotusymt:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | N/A |
---

- Documented the new **costmap** observation source (`nav2_msgs/msg/Costmap`).
- Added **⚠️ warning** about using **`CostmapSource`** (Collision Monitor normally **bypasses the costmap**).
- Updated the example **YAML** to include a `costmap` observation source entry.


SteveMacenski:
I think only thing is the migration guide can include an entry that this new feature was introduced!

PR #807, opened 2025-11-17
Consistently use ROS_DISTRO environment variable

benmccann:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | |
---

Updated line to use `$ROS_DISTRO` environment variable that's used in the following lines

### DIFF
--- a/tutorials/docs/navigation2_with_gps.rst
+++ b/tutorials/docs/navigation2_with_gps.rst
@@ -30,7 +30,7 @@ It is assumed ROS2 and Nav2 dependent packages are installed or built locally. A
 
    .. code-block:: bash
 
-      source /opt/ros/<ros2-distro>/setup.bash
+      source /opt/ros/$ROS_DISTRO/setup.bash
       sudo apt install ros-$ROS_DISTRO-nav2-minimal-tb3*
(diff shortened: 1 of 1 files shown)


SteveMacenski (approved):

tutorials/docs/navigation2_with_gps.rst
```diff
@@ -30,7 +30,7 @@ It is assumed ROS2 and Nav2 dependent packages are installed or built locally. A
 
    .. code-block:: bash
 
-      source /opt/ros/<ros2-distro>/setup.bash
+      source /opt/ros/$ROS_DISTRO/setup.bash
```
SteveMacenski, line 33: Has setup.bash been sourced in this session already? If not ROS_DISTRO is not set yet in this environment. It is set when sourcing that bash file happens, so you can use ROS_DISTRO afterwards, but not beforehand.

SteveMacenski: Thus, unfortunately, I think this PR should be  closed :(

benmccann: Ah! I was setting it manually and didn't realize that script did so. Today I learned! Thanks for filling me in!

PR #854, opened 2026-01-16
Refactor: change parameter naming style

AJedancov:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-navigation/navigation2/pull/5793 |
---

- Changed parameter names to reflect changes from PR in navigation2 project.
- Added a description of changes to the migration guide.
- Updated images for relevant BT.
- Updated XML files for BT.
- Changed the style of server names in some code examples.


behavior_trees/overview/detailed_behavior_tree_walkthrough.rst
```diff
@@ -47,12 +47,15 @@ BTs are primarily defined in XML. The tree shown above is represented in XML as
...
 
-    <root BTCPP_format="4" main_tree_to_execute="MainTree">
-        <BehaviorTree ID="MainTree">
+    <root BTCPP_format="4" main_tree_to_execute="NavigateToPoseWReplanningAndRecovery">
+        <BehaviorTree ID="NavigateToPoseWReplanningAndRecovery">
             <RecoveryNode number_of_retries="6" name="NavigateRecovery">
                 <PipelineSequence name="NavigateWithReplanning">
-                    <ControllerSelector selected_controller="{selected_controller}" default_controller="FollowPath" topic_name="controller_selector"/>
-                    <PlannerSelector selected_planner="{selected_planner}" default_planner="GridBased" topic_name="planner_selector"/>
+                    <ProgressCheckerSelector selected_progress_checker="{selected_progress_checker}" default_progress_checker="progress_checker" topic_name="progress_checker_selector"/>
```
SteveMacenski: Here and in general: This is going to be hard considering these docs touch all distributions :( changing the casing will make it so configuration files in Kilted and prior will not work with docs / BTs / yamls in the docs. 

I'd like to brain storm a way for how to handle this to make it more clear. For some like the configuration guide pages, I think we can pretty easily just have a note saying its `FollowPath` in Kilted and older and `follow_path` in Lyrical and newer. 

The BTs/YAMLs maybe just add a comment on each line `<!-- FollowPath in kilted and older -->` and `# FollowPath in kilted and older` anywhere the snake case is introduced?

Its alot of redundancy, but its the best I can think of right now without having different versions of docs which is something on our minds, but not something we have today

behavior_trees/trees/nav_to_pose_with_consistent_replanning_and_if_path_becomes_invalid.rst
```diff
@@ -26,20 +26,23 @@ Next, the recovery subtree will attempt the following recoveries: costmap cleari
...
-              <Fallback name="FallbackComputePathToPose">
-                <ReactiveSequence name="CheckIfNewPathNeeded">
+        <PipelineSequence name="NavigateWithReplanning">
+          <ControllerSelector selected_controller="{selected_controller}" default_controller="follow_path" topic_name="controller_selector"/>
+          <PlannerSelector selected_planner="{selected_planner}" default_planner="grid_based" topic_name="planner_selector"/>
+          <RateController hz="2.0">
+            <RecoveryNode number_of_retries="1" name="ComputePathToPose">
+              <Fallback>
+                <ReactiveSequence>
+                  <Inverter>
```
SteveMacenski: Note: this is a change from the previous version: Does this match that in Nav2 itself?

AJedancov: I took this behavior tree entirely from the Nav2 project:
https://github.com/ros-navigation/navigation2/blob/main/nav2_bt_navigator/behavior_trees/nav_to_pose_with_consistent_replanning_and_if_path_becomes_invalid.xml

I only changed the values for ports (follow_path and grid_based).
And the updated version now matches the description above, regarding replanning in case the path has not been updated for 10 seconds.

> In nominal execution, replanning can be triggered by an a invalid previous path, a new goal or if a new path has not been created for 10 seconds.

behavior_trees/overview/detailed_behavior_tree_walkthrough.rst
```diff
@@ -142,26 +145,33 @@ The XML of this subtree is as follows:
...
-        </RateController>
-        <RecoveryNode number_of_retries="1" name="FollowPath">
-            <FollowPath path="{path}" controller_id="{selected_controller}" error_code_id="{follow_path_error_code}" error_msg="{follow_path_error_msg}"/>
-            <Sequence>
-                <WouldAControllerRecoveryHelp error_code="{follow_path_error_code}"/>
-                <ClearEntireCostmap name="ClearLocalCostmap-Context" service_name="local_costmap/clear_entirely_local_costmap"/>
-            </Sequence>
-        </RecoveryNode>
-    </PipelineSequence>
+    <root BTCPP_format="4" main_tree_to_execute="NavigateToPoseWReplanningAndRecovery">
```
mini-1235: This should start with `<PipelineSequence>` instead of `<root>`

behavior_trees/overview/detailed_behavior_tree_walkthrough.rst
```diff
@@ -231,24 +241,28 @@ And the XML snippet:
...
-                <ClearEntireCostmap name="ClearLocalCostmap-Subtree" service_name="local_costmap/clear_entirely_local_costmap"/>
-                <ClearEntireCostmap name="ClearGlobalCostmap-Subtree" service_name="global_costmap/clear_entirely_global_costmap"/>
-                </Sequence>
-                <Spin spin_dist="1.57" error_code_id="{spin_error_code}" error_msg="{spin_error_msg}"/>
-                <Wait wait_duration="5.0" error_code_id="{wait_error_code}" error_msg="{wait_error_msg}"/>
-                <BackUp backup_dist="0.30" backup_speed="0.15" error_code_id="{backup_error_code}" error_msg="{backup_error_msg}"/>
-            </RoundRobin>
-        </ReactiveFallback>
-    </Sequence>
+    <root BTCPP_format="4" main_tree_to_execute="NavigateToPoseWReplanningAndRecovery">
```
mini-1235: This should start with `<Sequence>` instead of `<root>`

migration/Kilted.rst
```diff
@@ -5,6 +5,142 @@ Kilted to L-turtle
...
++-------------------------+--------------------------+----------------------------------------+
+| ComputePathThroughPoses | planner_id               | `GridBased` -> `grid_based`            |
++-------------------------+--------------------------+----------------------------------------+
+| SmoothPath              | smoother_id              | `RouteSmoother` -> `route_smoother`    |
++-------------------------+--------------------------+----------------------------------------+
+| ControllerSelector      | default_controller       | `FollowPath` -> `follow_path`          |
++-------------------------+--------------------------+----------------------------------------+
+| GoalCheckerSelector     | default_goal_checker     | `GoalChecker` -> `goal_checker`        |
++-------------------------+--------------------------+----------------------------------------+
+| ProgressCheckerSelector | default_progress_checker | `ProgressChecker` -> `progress_checker`|
```
mini-1235: Missing space after `progress_checker`

migration/Kilted.rst
```diff
@@ -5,6 +5,142 @@ Kilted to L-turtle
...
++-------------------------+--------------------------+----------------------------------------+
+| BT Action Node ID       | Input port               | Example                                |
++=========================+==========================+========================================+
+| FollowPath              | controller_id            | `FollowPath` -> `follow_path`          |
++-------------------------+--------------------------+----------------------------------------+
+| FollowPath              | path_handler_id          | `PathHandler` -> `path_handler`        |
++-------------------------+--------------------------+----------------------------------------+
+| FollowPath              | goal_checker_id          | `GoalChecker` -> `goal_checker`        |
++-------------------------+--------------------------+----------------------------------------+
+| FollowPath              | progress_checker_id      | `ProgressChecker` -> `progress_checker`|
```
mini-1235: Missing space after `progress_checker`

mini-1235 (changes requested):
Otherwise LGTM!

behavior_trees/overview/detailed_behavior_tree_walkthrough.rst
AJedancov: > Here and in general: This is going to be hard considering these docs touch all distributions :( changing the casing will make it so configuration files in Kilted and prior will not work with docs / BTs / yamls in the docs.

Oh yeah, that will be confusing in this case. 

> I'd like to brain storm a way for how to handle this to make it more clear.

Okay, I think we can leave a small note in brackets for the text.

> The BTs/YAMLs maybe just add a comment on each line  and # FollowPath in kilted and older anywhere the snake case is introduced?

For code blocks and images, I would consider an option using [tabs](https://sphinx-tabs.readthedocs.io/en/latest/) instead of comments.
I see several advantages here:
- Using `group-tab::`, we can switch all tabs synchronously when switching one on the same page. I've checked this works even with other pages in one folder (but not for entire project). An example of how it works can be seen on one of the pages of [ROS 2 documentation](https://docs.ros.org/en/rolling/Installation/Maintaining-a-Source-Checkout.html). 
- It's easier to copy from code or config examples. The content will include only relevant information.
- It will simplify the transition process when documentation for different versions becomes available. The only thing that needs to be done is to delete the tab for the corresponding version and leave its contents.

Here are some examples of what this might look like:
image

image   


I think for tables we can use additional column labeled `Default in Lyrical and newer` and `Default in Kilted and older`. 
image
Or we can apply a similar approach with tabs:
image
This can be useful for syncing between tabs with subsequent code examples.

Would you consider this approach?

behavior_trees/overview/detailed_behavior_tree_walkthrough.rst
AJedancov: Thanks for the review! Here I would like to discuss how make it better.
The thing is that above this BT example there is an image in which the parent node `NavigateToPoseWReplanningAndRecovery` is explicitly displayed.
Wouldn't it be better to leave the full code for the subtree so that it corresponds one to one with the image?
Or we can do the opposite and change the image to match the tree that starts with the `PipelineSequence`? And display it in the code somehow to emphasize that it is a part of existing configuration, for example using ellipsis points:
```xml
...
<PipelineSequence name="NavigateWithReplanning">
	<ProgressCheckerSelector selected_progress_checker="{selected_progress_checker}" default_progress_checker="progress_checker" topic_name="progress_checker_selector"/>
	<GoalCheckerSelector selected_goal_checker="{selected_goal_checker}" default_goal_checker="general_goal_checker" topic_name="goal_checker_selector"/>
	<PathHandlerSelector selected_path_handler="{selected_path_handler}" default_path_handler="path_handler" topic_name="path_handler_selector"/>
	<ControllerSelector selected_controller="{selected_controller}" default_controller="follow_path" topic_name="controller_selector"/>
	<PlannerSelector selected_planner="{selected_planner}" default_planner="grid_based" topic_name="planner_selector"/>
	<RateController hz="1.0">
		<RecoveryNode number_of_retries="1" name="ComputePathToPose">
			<ComputePathToPose goal="{goal}" path="{path}" planner_id="{selected_planner}" error_code_id="{compute_path_error_code}" error_msg="{compute_path_error_msg}"/>
			<Sequence>
				<WouldAPlannerRecoveryHelp error_code="{compute_path_error_code}"/>
				<ClearEntireCostmap name="ClearGlobalCostmap-Context" service_name="global_costmap/clear_entirely_global_costmap"/>
			</Sequence>
		</RecoveryNode>
	</RateController>
	<RecoveryNode number_of_retries="1" name="FollowPath">
		<FollowPath path="{path}" controller_id="{selected_controller}" error_code_id="{follow_path_error_code}" error_msg="{follow_path_error_msg}"/>
		<Sequence>
			<WouldAControllerRecoveryHelp error_code="{follow_path_error_code}"/>
			<ClearEntireCostmap name="ClearLocalCostmap-Context" service_name="local_costmap/clear_entirely_local_costmap"/>
		</Sequence>
	</RecoveryNode>
</PipelineSequence>
...
```

migration/Kilted.rst
AJedancov: Done in https://github.com/ros-navigation/docs.nav2.org/pull/854/commits/8d6ef31a3d2d57b3fa1038e68cd01e6f21cbbcc7

migration/Kilted.rst
AJedancov: Done

behavior_trees/overview/detailed_behavior_tree_walkthrough.rst
SteveMacenski: Oh I like that! Tabs for the BT XML and the additional column for the yaml parameter descriptions I think look best :-)

behavior_trees/overview/detailed_behavior_tree_walkthrough.rst
mini-1235: I think it's already clear from the previous paragraphs that this is part of the existing configuration. It would probably make more sense to me to stick with the original version that starts with `<PipelineSequence>`, and update the image to match that structure accordingly if possible

behavior_trees/overview/detailed_behavior_tree_walkthrough.rst
SteveMacenski: @AJedancov I changed my mind actually. Can you instead of doing the  additional column for the yaml parameter descriptions, have the entire configuration guide parameter section be under that tab? That would future proof for other additional parameter updates. Even if right now Jazzy/Kilted/Rolling are all the same, setting up that pattern in all of the configuration guide pages would help us for enabling multi-distribution docs moving forward.

Great idea!

behavior_trees/overview/detailed_behavior_tree_walkthrough.rst
AJedancov: I have updated the configuration and the corresponding images for them.

behavior_trees/overview/detailed_behavior_tree_walkthrough.rst
AJedancov: Done

behavior_trees/overview/detailed_behavior_tree_walkthrough.rst
AJedancov: > Can you instead of doing the additional column for the yaml parameter descriptions, have the entire configuration guide parameter section be under that tab?

Sure! Sounds good to me too.

AJedancov: I have prepared commit [b3037c0](https://github.com/ros-navigation/docs.nav2.org/pull/854/commits/b3037c02eb0ea29a006a2912bfe90df281409f63) with tabs. I just want to point out that in some places I have included adjacent paragraph into tab, for example in `behavior_trees/trees/nav_through_poses_recovery.rst`, since some configuration files of different distributions include different BT nodes and mentions of them. And in other places I inserted notes directly in the text, as discussed above.

In addition [939836f](https://github.com/ros-navigation/docs.nav2.org/pull/854/commits/939836f656f5c9c02c67efdec36905227668685f), I think it would be useful to add tabs for examples to the migration guide I prepared earlier.

configuration/packages/bt-plugins/actions/ComputePathThroughPoses.rst
```diff
@@ -10,61 +10,124 @@ The server address can be remapped using the ``server_name`` input port.
...
 Input Ports
 -----------
-:start:
 
-  ===================================== =======
-  Type                                  Default
-  ------------------------------------- -------
-  geometry_msgs::msg::PoseStamped         N/A
-  ===================================== =======
+.. tabs::
```
mini-1235, line 14: If I remember correctly, we don't use nav_msgs::msg::Goals in Jazzy and Older

configuration/packages/bt-plugins/actions/FollowPath.rst
```diff
@@ -9,84 +9,167 @@ The server address can be remapped using the ``server_name`` input port.
...
+
+    :path_handler_id:
+
+      ====== =======
+      Type   Default
+      ------ -------
+      string N/A
+      ====== =======
+
+      Description
```
mini-1235: The path handler does not exist in Kilted and older

configuration/packages/configuring-controller-server.rst
```diff
@@ -14,286 +14,573 @@ It also hosts the local costmap.
...
+    :search_window:
+
+      ============== =============================
+      Type           Default
+      -------------- -----------------------------
+      double           2.0
+      ============== =============================
+
+      Description
+        How far (in meters) along the path the searching algorithm will look for the closest point.
```
mini-1235: Path handler and search window only exist in Lyrical and newer

configuration/packages/configuring-planner-server.rst
```diff
@@ -14,118 +14,246 @@ It also hosts the global costmap.
...
+            planner_server:
+              ros__parameters:
+                planner_plugins: ["GridBased"]
+                GridBased:
+                  plugin: "nav2_navfn_planner::NavfnPlanner" # In Iron and older versions, "/" was used instead of "::"
+        ..
 
-  Description
-    The lifecycle node bond mechanism publishing period (on the /bond topic). Disabled if inferior or equal to 0.0.
+    :allow_partial_planning:
```
mini-1235: allow partial planning only exist in lyrical and newer

configuration/packages/configuring-regulated-pp.rst
```diff
@@ -427,58 +427,121 @@ Regulated Pure Pursuit Parameters
...
+            regulated_linear_scaling_min_radius: 0.9
+            regulated_linear_scaling_min_speed: 0.25
+            use_rotate_to_heading: true
+            allow_reversing: false
+            rotate_to_heading_min_angle: 0.785
+            min_distance_to_obstacle: 0.0
+            stateful: true
+            use_dynamic_window: false
+
+  .. group-tab:: Kilted and older
```
mini-1235, line 851: Kilted and older should be:
```
plugin: "nav2_regulated_pure_pursuit_controller::RegulatedPurePursuitController"
        desired_linear_vel: 0.5
        lookahead_dist: 0.6
        min_lookahead_dist: 0.3
        max_lookahead_dist: 0.9
        lookahead_time: 1.5
        rotate_to_heading_angular_vel: 1.8
        use_velocity_scaled_lookahead_dist: false
        min_approach_linear_velocity: 0.05
        approach_velocity_scaling_dist: 0.6
        use_collision_detection: true
        max_allowed_time_to_collision_up_to_carrot: 1.0
        use_regulated_linear_velocity_scaling: true
        use_fixed_curvature_lookahead: false
        curvature_lookahead_dist: 0.25
        use_cost_regulated_linear_velocity_scaling: false
        cost_scaling_dist: 0.3
        cost_scaling_gain: 1.0
        regulated_linear_scaling_min_radius: 0.9
        regulated_linear_scaling_min_speed: 0.25
        use_rotate_to_heading: true
        allow_reversing: false
        rotate_to_heading_min_angle: 0.785
        max_angular_accel: 3.2
        stateful: true
```

configuration/packages/configuring-rotation-shim-controller.rst
```diff
@@ -153,38 +153,81 @@ Rotation Shim Controller Parameters
...
+            max_angular_accel: 3.2
+            simulate_ahead_time: 1.0
+            rotate_to_goal_heading: false
+
+            # Primary controller params can be placed here below
+            primary_controller:
+              plugin: "nav2_regulated_pure_pursuit_controller::RegulatedPurePursuitController"
+              # ...
+
+  .. group-tab:: Kilted and older
```
mini-1235, line 327: Kilted and older should be:
```
FollowPath:
        plugin: "nav2_rotation_shim_controller::RotationShimController"
        primary_controller: "nav2_regulated_pure_pursuit_controller::RegulatedPurePursuitController"
        angular_dist_threshold: 0.785
        forward_sampling_distance: 0.5
        angular_disengage_threshold: 0.3925
        rotate_to_heading_angular_vel: 1.8
        max_angular_accel: 3.2
        simulate_ahead_time: 1.0
        rotate_to_goal_heading: false
        # Primary controller params can be placed here below
        # ...
````

without the primary controller namespace

File: configuration/packages/nav2_controller-plugins/feasible_path_handler.rst [hunk 13]
```diff
@@ -102,14 +102,33 @@ Parameters
...
-      enforce_path_inversion: True
-      enforce_path_rotation: False
-      inversion_xy_tolerance: 0.2
-      inversion_yaw_tolerance: 0.4
-      minimum_rotation_angle: 0.785
-      reject_unit_path: False
+
+.. tabs::
+
+  .. group-tab:: Lyrical and newer
```
mini-1235: Path handler does not exist in Kilted and older

tutorials/docs/using_shim_controller.rst
```diff
@@ -62,49 +66,108 @@ As such, its configuration looks very similar to that of any other plugin. In th
...
+      controller_server:
+        ros__parameters:
+          controller_frequency: 20.0
+          min_x_velocity_threshold: 0.001
+          min_y_velocity_threshold: 0.5
+          min_theta_velocity_threshold: 0.001
+          controller_plugins: ["follow_path"]
+          follow_path:
+            plugin: "nav2_rotation_shim_controller::RotationShimController"
+            primary_controller: "dwb_core::DWBLocalPlanner"
```
mini-1235: Please update according to https://github.com/ros-navigation/docs.nav2.org/pull/800/files, it looks like I forgot to update the tutorial in that PR :smile:

mini-1235 (changes requested):
Largely looks good to me. I left a few comments regarding parameters/types/plugins that don't exist in older distributions

configuration/packages/bt-plugins/actions/ComputePathThroughPoses.rst
AJedancov: You are right, in Jazzy is `geometry_msgs::msg::PoseStamped[]` instead. The situation is that Kilted already uses `nav_msgs::msg::Goals`. 
One more tab? :)

AJedancov: Or add a comment similar to this one:
```
collision_monitor:
	ros__parameters:
	  ...
	  enable_stamped_cmd_vel: True  # False for Jazzy or older
```

mini-1235: Yeah, I think this is one downside of using tabs to handle multi-version. Since we are usually referring to a range of distributions rather than a specific one, adding a tab for every distro doesn't feel right to me either

Thoughts? @SteveMacenski 

Maybe we should just remove the tabs and use comment instead as @AJedancov suggested in this case?

SteveMacenski: See https://github.com/ros-navigation/docs.nav2.org/pull/854#issuecomment-3801611166

configuration/packages/nav2_controller-plugins/feasible_path_handler.rst
AJedancov: What is better for pages like this that will appear in the new distribution only?
- Wrap the parameters/example into a single tab with the name of the new version. At least informative and will serve as a basis for future updates.
- Insert a Note/Warning section at the very top of the page with info about belonging to the new distribution.

mini-1235:
> These tabs (no comments, only relevant information)

> Would you consider removing all comments that relate to older versions from the Lyrical tab and leaving them only in the second tab? To stick to the idea I mentioned in the previous comment.

Can you provide an example?

> A few more thoughts on the upcoming warnings:
I think the insertion of them might require a separate PR that will address the problem on a broader scale.
Potential discrepancies in information between versions may also appear beyond the current topic. It's good that we were able to see them now and will try to eliminate, but as I already mentioned, they may arise in other places as well. Because at the moment I have only used tabs for those pages that are targeted at the relevant Nav2 PR.

If we are adding this in every page, this should be a few lines to the .html file:
https://github.com/ros2/ros2_documentation/blob/ce4c592d723d5f1602ee9a855185cfe5fa6c7fcb/source/_templates/page.html#L23-L41
but I am OK to deal with this in a separate PR

Also, please avoid force push, as it makes the review difficult given the size of this PR :)

mini-1235:
Yes! I think the comments can be removed in the lyrical and newer tab

SteveMacenski:
>  I would suggest focusing now on tabs for Lyrical and Kilted distributions:

Longer term, I think we should keep them targeted to a specific distribution, not grouped together. But I agree that for now Rolling (Lyrical) and then 'Kilted and Older' is fine considering that everything in Kilted and Older will already not up to date. As Kilted gets updated we can add increasing depth of older distributions as part of that work (separate of this PR). 

Only actionable note is that I'd like to use the name "Rolling" since Lyrical doesn't exist. Either way when Lyrical is cut, we'll have to add a new tab (either for the new head for Rolling, or for taking the current content in rolling and calling a tab before it Lyrical - same difference).

> I agree that we should limit this to two tabs for now. Instead of nesting warnings inside the tabs themselves, I think it would be clearer to place a warning block at the top of each page, similar to the ROS 2 documentation.

My suggestion is that it should be the first thing in the tab for any tabs that are not up-to-date (Kilted and older, for now). That way it only appears when you click on a bad tab. Every time we migrate a tab for a page, we can remove this warning to signal that its 'good info'. It would help us know which are updated and which are not as we go through that process & which tabs of which pages can be trusted by the readers. I'd rather this not on the top of each page for that reason, does that make sense? 

> Personally, I don't t see tabs as a long-term solution for multi-version docs, but it's a reasonable stopgap for now

I'm curious to hear why's that? I figure this is actually easier than maintaining separate branches which would involve additional PRs for docs for each backport. If tabs aren't the long term solution, I'd actually rather not do them at all here since that adds complexity to manage / revert later + alot of work to get Kilted/Older populated which would otherwise be wasted engineering cycles. Merging this involves some major time commitment in the nearterm future to populate (probably will try to recruit some community contributors + myself). 

I honestly have a hard time thinking of any 'perfect' solution to multidistro docs, which is why we don't have it as of the time of writing :/  Other than (a) auto generating them or (b) having a tool which extracts all parameters and validates they're on the pages they should be for each distribution or else fail the build, I'm lacking a great solution which makes sure things don't constantly fall through the cracks. Of these two, I'd personally favor (B) because (A) would involve having all comments inline in the code which would add alot of bloat. (B) makes it clear that there's a problem for immediate resolution, but also can fail on people's unrelated PRs adding burdens to them to fix things out of the scope of their work. Plus proactive PRs would add parameters not yet in those branches, so there's that.

I suppose some hybrid could be maintaining Rolling and have only the released distributions auto generated. We take the descriptions from the Rolling entries but populate the in-code default values specific to each distro. For parameters that are unique to a particular distro, we'd have to maintain a separate description list file it can pull from, but those are few and far between. Hardly a huge issue. 

But this is trading fairly manual parameter validation for up-front development cost in these tools. **I'm open to any/all suggestions.** Other options we've talked about in the past using tags and backport branches seemed too fragile to rely on. 

As far as I suggest:
1. We need to agree with how we plan to handle multi-distro docs first before we can know what to do
2. Then we can implement the current state as the "Older" docs + a warning and your changes for "Newer" that are up to date tracking Rolling
3. Have open action items to properly populate the "Older" docs 

It seems like we're blocked by (1) before we can merge this work since its such a breaking change all over the place. I know we stepped into quite the mess here that was not the original point. @AJedancov I don't want to keep piling on / make this death by a thousand cuts :(

AJedancov:
> My suggestion is that it should be the first thing in the tab for any tabs that are not up-to-date (Kilted and older, for now). That way it only appears when you click on a bad tab. Every time we migrate a tab for a page, we can remove this warning to signal that its 'good info'.

Tracking the progress of changes will be an advantage here, I agree. 

> I'd rather this not on the top of each page for that reason, does that make sense?

We can probably try using the combination again:
- Pages **with tabs** will include warning in Kilted and older tab.
- Pages **without tabs** will include warning on the top of page (for the rest configuration guides pages).

In each case users will be warned about possible version differences. And for us there will be a progress indicator that cover all places.

But we can postpone the discussion of these issues to the next PR after this. Perhaps other points of implementation or ideas will arise.

SteveMacenski:
> Especially for BT configuration xml and yaml files

I agree, with the declarative input/output ports, this should be easy, except when they derive from a base class. We'd need to check for that too.

>  And by the way, this could partially reduce the amount of information on the page that include tabs (regarding the topic of support and reviews). I've already tried to find a similar solution, but for now I've left text as before, so as not to stray too far from the main idea, and to leave this option for optimization for the future.

In what way?

>      We can probably try using the combination again: (1) Pages with tabs will include warning in Kilted and older tab. (2) Pages without tabs will include warning on the top of page (for the rest configuration guides pages).

I thought though perhaps the answer is not to use tables but use the multidistro support - which would resolve this a more scalable way towards our long-term goals?

Before you keep cranking this out, I think we should figure the direction. I think keeping this cleanly **only** you changes for now would be good and not trying to fix other distribution issues until we know what we're doing. Keep those changes and/or commits handy though for later.

mini-1235:
Hi @AJedancov, what's the current status of this PR?
I have another PR that's currently blocked by it https://github.com/ros-navigation/navigation2/pull/5876, mainly related to the detailed behavior tree walkthrough.
You're doing great work here, and I would like to avoid having you rebasing multiple times because of my PR :)

SteveMacenski:
Closing with previous PR

PR #900, opened 2026-04-10
fix dwb linear granularity default

EricoMeger:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | (N/A) |
| Does this PR contain AI-generated software? | (No) |
---

### DIFF
--- a/configuration/packages/dwb-plugins/standard_traj_generator.rst
+++ b/configuration/packages/dwb-plugins/standard_traj_generator.rst
@@ -46,7 +46,7 @@ Parameters
   ====== =======
   Type   Default
   ------ -------
-  double 0.5
+  double 0.05
   ====== =======
 
     Description


configuration/packages/dwb-plugins/standard_traj_generator.rst
```diff
@@ -46,7 +46,7 @@ Parameters
   ====== =======
   Type   Default
   ------ -------
-  double 0.5
+  double 0.05
```
SteveMacenski, line 49: Seems to me to be correct: https://github.com/ros-navigation/navigation2/blob/a65638d58e95bb71ab33f5441ec7022c4c7f984b/nav2_dwb_controller/dwb_plugins/src/standard_traj_generator.cpp#L72

SteveMacenski (changes requested):

configuration/packages/dwb-plugins/standard_traj_generator.rst
EricoMeger: I think I got kinda biased from my DWA experience, I should have double-checked to see if it was actually correct.

That said, this value seems a bit strange to me, I think this would be way too coarse for the trajectory samples. Can't really say its a typo though, since looking at the git history 0.5 is the default value since the initial implementation in 2018

I also don't think it would be right to change the config example, since it seems like DWB historically has been configured using 0.05 instead. See Humble, for example: https://github.com/ros-navigation/navigation2/blob/e9caa4287b99c17f4ff39fbf2315d4d82e78da93/nav2_bringup/params/nav2_params.yaml#L164

I'll be closing this one then, thanks for taking a look!

PR #905, opened 2026-04-21
Update documentation for pluginized motion models

Alessio-Parmeggiani:
---

## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | [nav2/6076](https://github.com/ros-navigation/navigation2/pull/6076) |
| Does this PR contain AI-generated software? | No|
---

Updated the documentation to account for changes added in this [PR](https://github.com/ros-navigation/navigation2/pull/6076) that add plugins for motion models in MPPI controller.

Requires the PR in nav2 repository to be merged before merging this.

### DIFF
--- a/configuration/packages/configuring-mppic.rst
+++ b/configuration/packages/configuring-mppic.rst
@@ -53,17 +53,6 @@ MPPI Parameters
     The plugin to use for the motion model constraints of the MPPI planner. 
     Supported motion model plugins include "mppi::DiffDriveMotionModel", "mppi::OmniMotionModel", and "mppi::AckermannMotionModel" for differential drive, omnidirectional, and Ackermann robots respectively.
 
-:``<motion_model>``.min_turning_r:
-
-  ============== ===========================
-  Type           Default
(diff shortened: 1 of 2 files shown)


configuration/packages/configuring-mppic.rst
```diff
@@ -374,19 +397,6 @@ Trajectory Visualization
...
-:min_turning_r:
-
-  ============== ===========================
-  Type           Default
-  -------------- ---------------------------
-  double         0.2
-  ============== ===========================
-
-  Description
-    The minimum turning radius possible for the vehicle platform (m).
```
SteveMacenski, line 402: You just want to add this back here and rename the section for the ackermann plugin name? Adding that in the generic arameters isn't correct I don't think. I think this was already well organized. The "Ackermann Motion Model" header should just be renamed as the plugin's name

SteveMacenski (approved):

SteveMacenski:
Should this be removed from draft mode?

PR #954, opened 2026-08-25
Document controller transform staleness threshold

nirwester:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | https://github.com/ros-navigation/navigation2/issues/6320 |
---

Documents the new transform_staleness_threshold parameter in the controller server.

### DIFF
--- a/docs/configuration_and_development/configuration_guide/core_servers/controller_server/index.md
+++ b/docs/configuration_and_development/configuration_guide/core_servers/controller_server/index.md
@@ -20,6 +20,14 @@ Type: `double` Default: `0.3`
 
 :   The timeout value (seconds) for the costmap to be fully updated before a control effort can be computed.
 
+### **`transform_staleness_threshold`**
+
+Type: `double` Default: `0.0`
+
+:   Maximum allowed age (seconds) of the transform used to obtain the robot pose in the local costmap's global frame.
+    If the transform is older than this threshold, the controller server reports a transform error instead of computing a velocity command.
+    The check is enabled for positive values; values less than or equal to `0.0` disable it.
+
 ### **`use_realtime_priority`**
 
 Type: `bool` Default: `false`
@@ -264,6 +272,7 @@ controller_server:
   ros__parameters:
     controller_frequency: 20.0
     costmap_update_timeout: 0.3
+    transform_staleness_threshold: 1.0
     min_x_velocity_threshold: 0.001
     min_y_velocity_threshold: 0.5
     min_theta_velocity_threshold: 0.001


SteveMacenski (approved):

docs/configuration_and_development/configuration_guide/core_servers/controller_server/index.md
```diff
@@ -264,6 +272,7 @@ controller_server:
   ros__parameters:
     controller_frequency: 20.0
     costmap_update_timeout: 0.3
+    transform_staleness_threshold: 1.0
```
SteveMacenski:
```suggestion
    transform_staleness_threshold: 0.0
```

Please copy paste these to start off with, setting this to something that low is not a sane general default I don't think

SteveMacenski (commented):
Lets leave this open as we discuss next steps with your PR, which may include a utility and expanding the use of this param to other locales

docs/configuration_and_development/configuration_guide/core_servers/controller_server/index.md
nirwester: The default is 0, these are just suggestions/examples, right? It's the same default value as AMCL's transform_tolerance, so 1.0 or a tiny bit more is what I'd set this to unless dealing with a very laggy system. Why would we suggest to have no check? :)

SteveMacenski: I think increasing this then, I think 1.0 would trip sometimes and be an annoying ghost in the system for users. How about 3.0?

SteveMacenski (approved):

SteveMacenski (approved):
need the migration guide entry for this (plan on expanding a bullet pointed list over time with the servers that contain it now)

docs/configuration_and_development/migration_guides/Lyrical.md
```diff
@@ -1 +1,33 @@
 # Lyrical to M-Turtle { #lyrical-to-m-turtle }
+
+## Transform Staleness Checking and Time-Coherent Transformations
+
+The Following Nodes can now optionally detect stale transforms used to obtain the robot pose in the local costmap's global frame:
```
SteveMacenski:
```suggestion
The Following Nodes can now optionally detect stale transforms:
```

Please explain the "why" so they know why they might want to configure and use this feature

docs/configuration_and_development/migration_guides/Lyrical.md
```diff
@@ -1 +1,33 @@
...
+
+- Controller Server
+
+A new transform_staleness_threshold parameter specifies the maximum allowed age of this transform in seconds. Positive values enable the staleness check, while values less than or equal to 0.0 disable it. The default value is 0.0, so no configuration changes are required to retain the previous behavior with respect to stale-transform rejection.
+
+When the check is enabled and the transform is older than the configured threshold, the Nodes will report an error.
+
+Whenever possible the nodes will retrieve the required transformations once at the beginning of each execution cycle and reuses them / their timestamp across the rest of the computation. This ensures that components within an execution cycle operate using time-coherent information. Custom plugins performing related TF operations should use the timestamp supplied with these transformations / poses where appropriate, rather than independently requesting additional transformations.
+
```
SteveMacenski:
```suggestion
```

That's an implementation detail, not a migration detail

docs/configuration_and_development/migration_guides/Lyrical.md
```diff
@@ -1 +1,33 @@
...
+```
+
+Now:
+
+```cpp
+virtual bool check(
+  const geometry_msgs::msg::PoseStamped & current_pose) = 0;
+```
+
+Custom `ProgressChecker` plugins must update both the declaration and implementation of `check()` to use a `const geometry_msgs::msg::PoseStamped &`. No other behavioral changes are required for progress checker plugins as a result of this API change.
```
SteveMacenski:
```suggestion
```

SteveMacenski:
@nirwester please update for the scope of all your PRs 

I think we should also create a new document which is our "Synchronization Strategy" which talks about our design patterns for time synchronization:

* Sensor data is used and transformed as necessary at its provided capture time
* Computed results with timestamps like paths, controls, poses are used and transformed with their respective timestamps
* When a server or algorithm needs to find a current time and/or global pose transform, it does so & passes this stamp to be used for all operations downstream to ensure synchronization
* etc (though that should cover the bases, no? I might be missing a case...)

What I'd like to also know from that audit is what APIs we tend to use for each. If there are consistent APIs (minus some outliers that could be updated) then that would be amazing to also have listed to make it abundantly clear what is used when, and why. 

We have several transform / get pose APIs (https://github.com/ros-navigation/navigation2/blob/main/nav2_util/src/robot_utils.cpp) let alone the raw `lookupTransform()` / `transform()` API calls that would be good to wrap neatly up if possible. It would be my joy to clean up or provide extreme clarity of how and when to use the TF APIs (even better if they're all calling our wrappers). If we can get away from the wild-west use of TF, then it becomes much easier to control its uses in the future and make sure its always doing the right thing.

PR #977, opened 2026-09-23
Docs for Nav2 6410

### LINKED ISSUE #971: Update documentation for consolidated lifecycle manager
Follow-up to https://github.com/ros-navigation/navigation2/pull/6410

https://github.com/ros-navigation/navigation2/pull/6410 consolidates Nav2 lifecycle management under a single lifecycle manager in bringup_launch.py and removes the lifecycle managers previously created by the nested launch files.

As discussed in https://github.com/ros-navigation/navigation2/pull/6410, this changes the expected pattern for users launching Nav2 components or lifecycle nodes independently. The Rolling documentation should be updated to reflect the new lifecycle management structure and provide guidance for users adding their own lifecycle nodes.

----------------------------------

I did a quick look and of the 67 pages on that search I only think these matter:

* https://docs.nav2.org/rolling/tutorials/general_tutorials/adding_a_nav2_task_server/ (new task server instructions; to include the launch file stuff - basically just explain whatever we do)
* https://docs.nav2.org/rolling/tutorials/general_tutorials/isaac_perceptor/ (general note; not even sure if we need to update anything)
* https://docs.nav2.org/rolling/configuration_and_development/first_time_robot_setup_guide/footprint/setup_footprint/ (need to lifecycle manage)
* https://docs.nav2.org/rolling/tutorials/general_tutorials/navigation2_with_slam/navigation2_with_slam/ (need to lifecycle manage, same as above)
* https://docs.nav2.org/rolling/tutorials/general_tutorials/navigation2_on_real_turtlebot3/navigation2_on_real_turtlebot3/  (need to lifecycle manage, same as above)
* https://docs.nav2.org/rolling/configuration_and_development/first_time_robot_setup_guide/sensors/mapping_localization/   (need to lifecycle manage, same as above)
* https://docs.nav2.org/rolling/tutorials/general_tutorials/navigation2_with_speed_filter/navigation2_with_speed_filter/ and https://docs.nav2.org/rolling/tutorials/general_tutorials/navigation2_with_keepout_filter/navigation2_with_keepout_filter/ (update launch files with whatever the new pattern if we do here, so just a copy+paste)

So I see like ~4 unique things, 2 repeated in a few places

_Originally posted by @SteveMacenski in https://github.com/ros-navigation/navigation2/issues/6410#issuecomment-5501230971_

itsOsamasaid:
## Basic Info

| ------ | ----------- |
| Ticket(s) this addresses   | Fixes #971, follows ros-navigation/navigation2#6410 |
| Does this PR contain AI-generated software? | Yes. The drafts were written with AI help, disclosed in #971 and in the commit message. I checked each change against the merged code and tested the commands myself, see below. |

---

navigation2#6410 replaced the per launch file lifecycle managers with a single `lifecycle_manager_nav2` started by `bringup_launch.py`, and the nested launch files now expose `get_lifecycle_nodes(context)` instead of starting a manager of their own. This updates the pages that describe the old layout, following the plan agreed in #971.

* `navigation2_with_slam`, `setup_footprint`, `mapping_localization` and `isaac_perceptor` told the reader to run `navigation_launch.py` on its own, which now starts the servers without anything to configure or activate them. They now use `ros2 launch nav2_bringup bringup_launch.py use_localization:=False`, with a short note about where the lifecycle manager lives. All four pages get the map and the `map` to `odom` transform from somewhere else already.
* `navigation2_with_keepout_filter` and `navigation2_with_speed_filter` embed a copy of `keepout_zone_launch.py` / `speed_zone_launch.py` that still contained `lifecycle_manager_keepout_zone` / `lifecycle_manager_speed_zone`. Both blocks are replaced with the current upstream files and the surrounding text explains `get_lifecycle_nodes()`.
* `adding_a_nav2_task_server` built the node list by hand and named the manager `lifecycle_manager_navigation`. The `sensor_driver` example now uses `get_lifecycle_nodes()` inside an `OpaqueFunction` with `lifecycle_manager_nav2`.
* `navigation2_with_zone_parameter_filter` keeps its own manager, which still works, and gains one line pointing at the consolidated manager as an alternative.
* Adds a "Single Lifecycle Manager in Nav2 Bringup" entry to the Lyrical migration guide.

## Testing

`mkdocs build --strict` passes. The commands were run on `ghcr.io/ros-navigation/nav2_docker:rolling-nightly-standard`:

* `ros2 launch nav2_bringup navigation_launch.py` on its own: no errors, no lifecycle manager, and after a minute all eleven servers are still `unconfigured`.
* `ros2 launch nav2_bringup bringup_launch.py use_localization:=False`: every managed node reaches `active`, with no `amcl` and no `map_server`.
* The task server example: `node_names` comes out as `['sensor_driver', 'controller_server', ...]` and the manager configures and activates `sensor_driver` before the navigation servers.
* `navigation2_on_real_turtlebot3` needs no change: launching with `autostart:=False` and then calling `/lifecycle_manager_nav2/manage_nodes` with `STARTUP`, which is what the RViz Start button sends, brings the servers up.

## Notes for the reviewer

* Refreshing the two zone launch file blocks also brings in the `use_intra_process_comms` argument, which came from #5804 rather than #6410, so those diffs are a little larger than the manager change alone. Happy to trim them if you would rather keep the diff minimal.
* Still open from #971: with `use_localization:=False` alone, `bringup_launch.py` also starts the keepout and speed mask servers with an empty mask, and the costmaps keep logging `Filter mask was not received`. The old command never started those servers. Let me know if the four tutorial commands should also pass `use_keepout_zones:=False use_speed_zones:=False`, as the simulation launch files do, and I will update them.
* Since navigation2#6517 backported #6410 to Lyrical, this probably wants the `backport-lyrical` label.


docs/configuration_and_development/first_time_robot_setup_guide/footprint/setup_footprint.md
```diff
@@ -69,12 +69,16 @@ ros2 run tf2_ros static_transform_publisher 0 0 0 0 0 0 map odom
...
 
-Lastly, we will launch Nav2 using the `nav2_params.yaml` configuration file we just made and `navigation_launch.py`, the built-in launch file of `nav2_bringup`. Open a new terminal and execute the following:
+Lastly, we will launch Nav2 using the `nav2_params.yaml` configuration file we just made and `bringup_launch.py`, the built-in launch file of `nav2_bringup`. Since we are not using `nav2_amcl` or `nav2_map_server` in this guide, we set `use_localization:=False`. Open a new terminal and execute the following:
 
 ```shell
-ros2 launch nav2_bringup navigation_launch.py params_file:=<full/path/to/config/nav2_params.yaml>
+ros2 launch nav2_bringup bringup_launch.py use_localization:=False params_file:=<full/path/to/config/nav2_params.yaml>
 ```
 
+!!! note
```
mini-1235: I think this might be a bit too much detail, especially since this is the first time we mention the lifecycle manager or `navigation_launch.py` in the whole Markdown file (?)

docs/configuration_and_development/first_time_robot_setup_guide/sensors/mapping_localization.md
```diff
@@ -196,13 +196,17 @@ sudo apt install ros-$ROS_DISTRO-navigation2
...
 
 ```shell
-ros2 launch nav2_bringup navigation_launch.py use_sim_time:=true
+ros2 launch nav2_bringup bringup_launch.py use_localization:=False use_sim_time:=true
 ```
 
-Note that the parameters of the `nav2_costmap_2d` that we discussed in the previous subsection are included in the default parameters of `navigation_launch.py`. Aside from the `nav2_costmap_2d` parameters, it also contains parameters for the other nodes that are included in Nav2 implementation.
+!!! note
+
+    `bringup_launch.py` starts the lifecycle manager that brings the Nav2 servers up. Launching `navigation_launch.py` on its own leaves them unconfigured.
```
mini-1235: I think this might be a bit too much detail, especially since this is the first time we mention the lifecycle manager or navigation_launch.py in the whole Markdown file (?)

docs/configuration_and_development/migration_guides/Lyrical.md
```diff
@@ -18,6 +18,12 @@ The behavior plugins' parameters are now declared under the plugin's name from `
 
 Substitute your own plugin names if they differ from the defaults. Simply move the parameter under the plugin's declaration in your configuration.
 
+### Single Lifecycle Manager in Nav2 Bringup
+
+[PR #6410](https://github.com/ros-navigation/navigation2/pull/6410) combines the separate lifecycle managers for localization, navigation, SLAM, keepout zones, speed zones and the loopback simulator into a single `lifecycle_manager_nav2`, started by `bringup_launch.py`. This reduces Nav2's CPU use by about a quarter. The nested launch files in `nav2_bringup` no longer start a lifecycle manager of their own. Each has a `get_lifecycle_nodes(context)` function that returns its lifecycle node names, which `bringup_launch.py` collects into the `node_names` of the single manager.
```
mini-1235: I think we should mention that this was tested using the TB4 launch setup, which previously had four lifecycle managers, and that this PR consolidates them into a single lifecycle manager

docs/configuration_and_development/migration_guides/Lyrical.md
```diff
@@ -18,6 +18,12 @@ The behavior plugins' parameters are now declared under the plugin's name from `
...
+
+Launching a nested file by itself, such as `ros2 launch nav2_bringup navigation_launch.py`, now leaves the servers unconfigured. Launch `bringup_launch.py` instead and use its arguments to turn off what you do not need, for example `use_localization:=False` when SLAM or another source provides the map and the `map` to `odom` transform. Anything that used the old manager names, such as `lifecycle_manager_navigation/manage_nodes`, should use `lifecycle_manager_nav2` instead; the Nav2 RViz panel is already updated. Custom bringup launch files can build their `node_names` from the `get_lifecycle_nodes()` functions plus their own nodes, see the [task server tutorial][adding-a-new-nav2-task-server].
```
mini-1235, line 25: Before this PR, I was thinking about launching `navigation_launch.py` directly and then starting the lifecycle manager separately with `ros2 run ...` (Sorry, I don't remember the exact command). I think it would be useful to document this as an alternative so we don't always have to use `bringup_launch.py` with specific arguments

docs/tutorials/general_tutorials/isaac_perceptor/index.md
```diff
@@ -443,10 +443,14 @@ ros2 launch opennav_visual_nav_demo visual_nav_demo_launch.py
...
-    ros2 launch nav2_bringup navigation_launch.py
+    ros2 launch nav2_bringup bringup_launch.py use_localization:=False
     ```
 
-    to launch Nav2. NVIDIA provides its own VSLAM stack, so Nav2 is not launched by default.
+    to launch Nav2. NVIDIA provides its own VSLAM stack, so Nav2 is not launched by default, and `use_localization:=False` skips `nav2_amcl` and `nav2_map_server`.
+
+    !!! note
+
+        `bringup_launch.py` starts the lifecycle manager that brings the Nav2 servers up. Launching `navigation_launch.py` on its own leaves them unconfigured.
```
mini-1235: I think this might be a bit too much detail, especially since this is the first time we mention the lifecycle manager or navigation_launch.py in the whole Markdown file (?)

docs/tutorials/general_tutorials/navigation2_with_keepout_filter/navigation2_with_keepout_filter.md
```diff
@@ -74,7 +74,7 @@ Since filter mask image was created as a copy of main map, other fields of YAML-
...
 
-In order to enable Keepout Filter in your configuration, both servers should be enabled as a lifecycle nodes in Python launch-file. It is also possible to add them as Composition Nodes to your Navigation Component Container, which might look as follows:
+In order to enable Keepout Filter in your configuration, both servers should be enabled as a lifecycle nodes in Python launch-file. It is also possible to add them as Composition Nodes to your Navigation Component Container. Nav2 ships this as `keepout_zone_launch.py`, which `bringup_launch.py` includes when `use_keepout_zones` is `True`. It does not start a Lifecycle Manager of its own: `get_lifecycle_nodes()` returns the two node names and `bringup_launch.py` adds them to `lifecycle_manager_nav2`. In your own launch file, add both servers to the `node_names` of your Lifecycle Manager. The launch file might look as follows:
```
mini-1235: I think this might be a bit too much detail, especially since this is the first time we mention the lifecycle manager or keepout_zone_launch.py in the whole Markdown file (?)

docs/tutorials/general_tutorials/navigation2_with_slam/navigation2_with_slam.md
```diff
@@ -70,13 +70,17 @@ ros2 launch turtlebot3_bringup robot.launch.py
...
-Launch Navigation without nav2_amcl and nav2_map_server.
+Launch Navigation without nav2_amcl and nav2_map_server, which is what `use_localization:=False` does.
 It is assumed that the SLAM node(s) will publish to /map topic and provide the map->odom transform.
 
 ```bash
-ros2 launch nav2_bringup navigation_launch.py
+ros2 launch nav2_bringup bringup_launch.py use_localization:=False
 ```
 
+!!! note
```
mini-1235: I think this might be a bit too much detail, especially since this is the first time we mention the lifecycle manager or navigation_launch.py in the whole Markdown file (?)

mini-1235:
> Still open from https://github.com/ros-navigation/docs.nav2.org/issues/971: with use_localization:=False alone, bringup_launch.py also starts the keepout and speed mask servers with an empty mask, and the costmaps keep logging Filter mask was not received. The old command never started those servers. Let me know if the four tutorial commands should also pass use_keepout_zones:=False use_speed_zones:=False, as the simulation launch files do, and I will update them.

Yes, I think we should keep the old behavior unchanged, please pass the necessary arguments to the launch file so it works normally

Also please sign with DCO :)

docs/configuration_and_development/migration_guides/Lyrical.md
itsOsamasaid: Good idea, added it. The command is:

```bash
ros2 launch nav2_bringup navigation_launch.py
```

```bash
ros2 run nav2_lifecycle_manager lifecycle_manager --ros-args -r __node:=lifecycle_manager_nav2 \
  -p autostart:=true \
  -p node_names:="[controller_server, smoother_server, planner_server, route_server, behavior_server, \
  velocity_smoother, collision_monitor, bt_navigator, waypoint_follower, docking_server, following_server]"
```

I tested this on the rolling nightly image. The servers stay unconfigured after the launch file, and they all go active once the manager starts.

Two things I am not sure about:

- I named the manager `lifecycle_manager_nav2` with `-r __node:=` so the RViz panel still finds it. Without that it comes up as `lifecycle_manager` and the Startup button has nothing to talk to. Is that what you had in mind?
- The snippet hard codes the eleven node names, which will drift when a server is added. I added a line saying the names come from that launch file's `get_lifecycle_nodes()`, but let me know if you would rather keep it shorter or drop the list.

mini-1235: Oh, that is actually longer than I expected, but I think it is fine to keep all the node names, along with the correct lifecycle manager name

mini-1235:
@itsOsamasaid You haven't signed off with the DCO yet. Please check: https://github.com/ros-navigation/docs.nav2.org/pull/977/checks?check_run_id=107392444749

Other than the remaining open item above, I don't have much else to add. I will leave the final decision to @SteveMacenski

docs/configuration_and_development/migration_guides/Lyrical.md
itsOsamasaid: Kept it as is then. Added in 97df092.

